PR #18, opened 2022-04-10
Pass guild id to new column 'reportedToGuild'

Flip-Liquid:


alexkeating:
This looks good, I will push up a pr to fix the build issues

PR #54, opened 2022-06-09
ADD contract sync job

amrro:
This loads all the events from the contract via Subgraph and insert them into the db.

### DIFF
--- a/apps/contract-sync-job/project.json
+++ b/apps/contract-sync-job/project.json
@@ -29,9 +29,7 @@
     "serve": {
       "executor": "@nrwl/node:node",
       "options": {
-        "buildTarget": "contract-sync-job:build",
-        // "inspect": true,
-        "port": 55362
+        "buildTarget": "contract-sync-job:build"
       }
     },
     "lint": {
--- a/apps/contract-sync-job/project.json
+++ b/apps/contract-sync-job/project.json
@@ -29,7 +29,9 @@
     "serve": {
       "executor": "@nrwl/node:node",
       "options": {
-        "buildTarget": "contract-sync-job:build"
+        "buildTarget": "contract-sync-job:build",
+        // "inspect": true,
+        "port": 55362
       }
     },
     "lint": {
--- a/apps/contract-sync-job/src/main.ts
+++ b/apps/contract-sync-job/src/main.ts
@@ -1,22 +1,25 @@
 import { GovrnContract, NetworkConfig } from '@govrn/govrn-contract-client';
 import { GovrnGraphClient } from '@govrn/govrn-subgraph-client';
-import { GovrnProtocol, SortOrder } from '@govrn/protocol-client';
+import {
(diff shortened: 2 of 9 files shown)


apps/contract-sync-job/project.json
```diff
@@ -29,7 +29,9 @@
     "serve": {
       "executor": "@nrwl/node:node",
       "options": {
-        "buildTarget": "contract-sync-job:build"
+        "buildTarget": "contract-sync-job:build",
+        // "inspect": true,
```
alexkeating, line 33: This should probably be removed

apps/contract-sync-job/src/main.ts
```diff
@@ -1 +1,183 @@
...
+const jobName = 'contract-sync-job';
+
+const contractAddress = '0xefad47a6888762b666727cae19d5e502f0a68f43';
+const networkConfig: NetworkConfig = {
+  address: contractAddress,
+  chainId: 2,
+};
+
+const provider = new ethers.providers.JsonRpcProvider(
+  'https://rinkeby.infura.io/v3/cda40b6016f04aa48da096c096068718'
```
alexkeating, line 18: It would probably be good to make both this and the contract address config variables

apps/contract-sync-job/src/main.ts
```diff
@@ -1 +1,183 @@
...
+  // Query existing user.
+  const user = await govrn.user.list({
+    where: {
+      address: { equals: data.address },
+    },
+    first: 1,
+  });
+
+  // Returns the id, if user exists.
+  if (user.length != 0) {
```
alexkeating, line 35: We should be using !== here instead

apps/contract-sync-job/src/main.ts
```diff
@@ -1 +1,183 @@
...
+  }
+
+  // Insert new user into db, if user doesn't exist.
+  const newUser = await govrn.user.create({
+    data: {
+      address: data.address,
+      chain_type: {
+        connectOrCreate: {
+          where: { name: 'example2' },
+          create: { name: 'example2' },
```
alexkeating, line 46: Can we set this to rinkeby in this case? With a comment to pass in the appropriate chain?

apps/contract-sync-job/src/main.ts
```diff
@@ -1 +1,183 @@
...
+      console.log(`:: Processing Contribution: ${event.id} => ${contr.name}`);
+
+      const userId = await getOrInsertUser(govrn, {
+        address: event.address,
+      });
+
+      return {
+        name: contr.name,
+        status_id: 2,
+        activity_type_id: 1,
```
alexkeating, line 118: We have to fetch or create the activity type

alexkeating (commented):
Looking good, almost we need just a couple changes

alexkeating (approved):

PR #79, opened 2022-07-05
Update contract sync job to load off-chain details

amrro:
Update Contract Sync Job to use the off-chain ipfs uri to load data like: `name`, `details`& `proof` in order to be inserted in the `Contributions` table.

-----
Related Linear Ticket: [PRO-257](https://linear.app/govrn/issue/PRO-257/drop-tables-before-running-contract-sync-job)

### DIFF
--- a/apps/contract-sync-job/src/main.ts
+++ b/apps/contract-sync-job/src/main.ts
@@ -1,3 +1,4 @@
+import fetch from 'node-fetch';
 import { GovrnContract, NetworkConfig } from '@govrn/govrn-contract-client';
 import { GovrnGraphClient } from '@govrn/govrn-subgraph-client';
 import {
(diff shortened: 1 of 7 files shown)


alexkeating (commented):
Looks good, the Authorization header needs to be an env variable for security purposes

apps/contract-sync-job/src/main.ts
```diff
@@ -11,6 +11,7 @@ import { GraphQLClient } from 'graphql-request';
 
 const protcolUrl = process.env.PROTOCOL_URL;
 const SUBGRAPH_ENDPOINT = process.env.SUBGRAPH_URL;
+const KEVIN_MALONE_TOKEN = process.env.KEVIN_MALONE_TOKEN;
```
alexkeating, line 14: We probably shouldn't be reusing tokens across different jobs. We should create a new token in protocol api called CONTRACT_SYNC_TOKEN and use that in this job

apps/contract-sync-job/src/main.ts
```diff
@@ -112,8 +113,9 @@ const main = async () => {
 
   const graphQLClient = new GraphQLClient(SUBGRAPH_ENDPOINT);
   const govrn = new GovrnProtocol(protcolUrl, null, {
-    Authorization: 'Flutes',
+    Authorization: KEVIN_MALONE_TOKEN,
   });
+  console.dir(KEVIN_MALONE_TOKEN);
```
alexkeating, line 118: We probably want to remove this

PR #85, opened 2022-07-11
Feature/add all daos select

alexkeating:


apps/protocol-api/src/main.ts
```diff
@@ -51,7 +51,8 @@ const permissions = shield(
       activityTypes: or(isAuthenticated, hasToken),
       attestations: isAuthenticated,
       getUser: isAuthenticated,
-      guild: hasToken,
+      guild: isAuthenticated,
```
alexkeating, line 54: We should use an or here because this will probably break one of the jobs

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -274,6 +296,31 @@ export class ContributionCustomResolver {
...
+
+      await prisma.contribution.update({
+        data: {
+          ...(args.data.guildId && {
+            guilds: {
+              delete: [{
+                guild_id_contribution_id: {
+                  contribution_id: args.data.contributionId,
+                  guild_id: args.data.currentGuildId,
+                  // guild_id: currentContribution.guilds[0].guild.id
```
alexkeating, line 312: Probably good to remove

jonathanprozzi: Cool - removed this and also the now unused `findUnique` above it

PR #89, opened 2022-07-18
fix/update-username (PRO-254) (#87)

alexkeating:
* Profile update now rehydrates with new name


jonathanprozzi:
Looks good but I have a PR up for a fix for an edge case (leaving DAO optional on initial Report Form) -- https://github.com/Govrn-HQ/govrn-monorepo/pull/90

PR #93, opened 2022-07-19
Fix pages horizontal scroll

amrro:


apps/protocol-frontend/src/components/Sidebar.tsx
```diff
@@ -5,15 +5,14 @@ import {
...
   FiPlusSquare,
-  FiDisc,
   FiTwitter,
+  FiUsers,
 } from 'react-icons/fi';
 import Logo from './Logo';
 import NavButton from './NavButton';
-import NavLink from './NavLink';
 import ConnectWallet from './ConnectWallet';
+import { FaDiscord } from 'react-icons/all';
```
jonathanprozzi, line 15: Nice!

.nvmrc
```diff
@@ -0,0 +1 @@
+v14.19.3
```
jonathanprozzi, line 1: I think Keating and I may be using v16.13.1 -- we should all sync up on this

apps/protocol-frontend/README.md
```diff
@@ -0,0 +1,11 @@
+## Summary
+The main Frontend app for the Govrn Protocol. 
+
+## Nx Plugin
+This App is generated using [`@nxext/vite`](https://nxext.dev/docs/vite/overview.html). Check all targets at [`project.json`](./project.json)
+
+
+## `.env`
+
+- `VITE_PROTOCOL_URL`: e.g. http://localhost:4000/graphql
```
jonathanprozzi, line 10: Awesome! Thanks for starting the README

apps/protocol-frontend/src/components/ContributionsTable.tsx
```diff
@@ -1,25 +1,26 @@
-import { useState, useEffect, useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import {
-  Table,
+  Box,
+  chakra,
   HStack,
   IconButton,
+  Stack,
+  Table,
```
jonathanprozzi, line 8: Looks like <Tooltip/> needs to be added again. I think it was in before but not sure how it was removed -- don't think it was in these changes though but we can import it from chakra

jonathanprozzi (commented):
We'll need to add the <Tooltip/> import back in (not sure where it went)  It's causing the build to fail -- also I think we may all want to sync on our vscode settings -- are you using a particular setting to auto-sort imports? I like it -- I'll start using it as well so we can minimize changes from that

jonathanprozzi:
> > are you using a particular setting to auto-sort imports?
> 
> I am using Webstorm and integrating to run prettier and eslint on every save. I am not sure if WebStorm does some extra work. AFAIR vscode has a similar command. [more](https://stackoverflow.com/a/67838108/2489403)
> 
> But we can sync on that and allow linters to do the job. Also, I am planning to reformat the project remove unused imports and so on. But waiting for time like weekend where the project is cold in order to reduce conflicts.
> 
> I fixed the missing import and updated the .nvmrc. It was missed in the merge.

Sounds good! I'm running the lint and format on save in VSCode so not sure why my auto-sort is different for imports. I'll figure that out -- there are definitely some unused imports to clean up so that'll be great. I removed some here and there, but having a dedicated sweep is a great idea.

alexkeating:
Looks good, but I think we should remove the scroll on the sidebar. What are your thoughts?

Screenshot from 2022-07-19 12-52-20
r

amrro:
@alexkeating removed here [461b407](https://github.com/Govrn-HQ/govrn-monorepo/pull/93/commits/461b407a3bf0b0163f93abae3dcecbb80c94ecf7)

alexkeating:
Oh I think there was a miscommunication I was talking about this scrollbar on the sidebar. I think we should remove that one not the table scrollbars

Screenshot from 2022-07-19 21-53-21

jonathanprozzi:
> Oh I think there was a miscommunication I was talking about this scrollbar on the sidebar. I think we should remove that one not the table scrollbars
> 
> 
> 
> Screenshot from 2022-07-19 21-53-21
> 
> 
> 
> 

Yeah we'll want to remove this since it wasn't happening before. 👍

PR #96, opened 2022-07-19
Feature/more subgraph config

alexkeating:


alexkeating:
It looks like there are some changes in master that are missing from staging. This will be good to get things in sync

jonathanprozzi (approved):
Looks good!

PR #98, opened 2022-07-19
Make proof optional

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-286/proof-null-on-migrated-data

## Description

Make proof optional on update


jonathanprozzi (approved):
🚀

PR #99, opened 2022-07-19
Redirect x enhance

amrro:
## Linear Ticket
[PRO-209 - Improve redirect UX](https://linear.app/govrn/issue/PRO-209/improve-redirect-ux#update-c36aad14=)

## Description
Enhance UX when user is not active yet (on waiting list).

### DIFF
--- a/apps/protocol-frontend/src/components/HomeShell.tsx
+++ b/apps/protocol-frontend/src/components/HomeShell.tsx
@@ -8,6 +8,7 @@ import CreateUserForm from './CreateUserForm';
 
 import { useUser } from '../contexts/UserContext';
 import CreateWaitlistUserForm from './CreateWaitlistUserForm';
+import { GOVRN_MOTTO } from '../utils/constants';
 
 const HomeShell = () => {
   const { isConnected } = useWallet();
@@ -165,7 +166,7 @@ const HomeShell = () => {
           </span>
         </PageHeading>
         <Text color="gray.800" paddingBottom={4}>
-          Anything that governs you, you should be able to govern.
+          {GOVRN_MOTTO}
         </Text>
         {isConnected && isAuthenticated ? (
           <NewUserFlow />
--- a/apps/protocol-frontend/src/components/HomeShell.tsx
+++ b/apps/protocol-frontend/src/components/HomeShell.tsx
@@ -1,7 +1,14 @@
 import { useEffect, useState } from 'react';
 import { Link } from 'react-router-dom';
 import { useWallet } from '@raidguild/quiver';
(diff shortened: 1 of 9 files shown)


apps/protocol-frontend/src/components/HomeShell.tsx
```diff
@@ -115,11 +124,12 @@ const HomeShell = () => {
                   >
                     {userDataByAddress?.name}
                   </Text>
-                  . We'll reach out as soon as we open up more spots!
+                  . <b>We'll reach out as soon as we open up more spots!</b>
```
jonathanprozzi, line 127: We may want to use <strong> instead for this for screenreader emphasis-- this is super minor though. Good idea to call this text out

apps/protocol-frontend/src/components/HomeShell.tsx
```diff
@@ -1,13 +1,22 @@
...
+  Box,
+  Icon,
+} from '@chakra-ui/react';
 import ConnectWallet from '../components/ConnectWallet';
 import PageHeading from './PageHeading';
 import CreateUserForm from './CreateUserForm';
 
 import { useUser } from '../contexts/UserContext';
 import CreateWaitlistUserForm from './CreateWaitlistUserForm';
+import { GOVRN_MOTTO } from '../utils/constants';
```
jonathanprozzi, line 18: Nice. I have a few other things we can add to this now that it is here (constants.ts)

apps/protocol-frontend/src/pages/Redirect.tsx
```diff
@@ -0,0 +1,28 @@
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useState } from 'react';
+import { Flex, Spinner } from '@chakra-ui/react';
+
+const TIMEOUT = 3_000;
```
jonathanprozzi, line 5: We can perhaps drop this if we think it's a bit too long.

@alexkeating what do you think?

jonathanprozzi (approved):
Good improvements. I have a few comments in here but they're pretty minor. Mainly around the timing and 1 other small thing with using <strong> over <b>

alexkeating:
Looks great @amrro, I lowered the timeout to 300ms and @jonathanprozzi  it looks like redirects were disabled for the contribution route. I reenabled authentication. Do you see any problems with that

jonathanprozzi:
> Looks great @amrro, I lowered the timeout to 300ms and @jonathanprozzi  it looks like redirects were disabled for the contribution route. I reenabled authentication. Do you see any problems with that

Oh weird! Great catch. That was unintentional  

This is a great ux improvement @amrro 🎉

PR #103, opened 2022-07-21
This is required- activityType

NKoech123:
**PRO-241: Activity Type select throws an error**

## Description
Added 'activityType' field in yup's validation schema (in protocolfrontend/src/utils/validation.ts).
Achieved goal - no longer throws an error on submit when Activity Type field is empty. Instead, message 'This is required'
                           shows below the input box.
**TBD:**     pair up with @jonathanprozzi to add remaining  functions

## Screencaptures

**Before:**  Message below was being displayed

Screen Shot 2022-07-20 at 5 07 18 PM

**After:** The message is more User friendly

Screen Shot 2022-07-20 at 3 54 20 PM

### DIFF
--- a/apps/protocol-frontend/src/contexts/UserContext.tsx
+++ b/apps/protocol-frontend/src/contexts/UserContext.tsx
@@ -270,6 +270,7 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
         isClosable: true,
         position: 'top-right',
       });
+      console.log(values);
       getUserActivityTypes();
       getUserContributions();
       getDaoContributions();
--- a/apps/protocol-frontend/src/contexts/UserContext.tsx
+++ b/apps/protocol-frontend/src/contexts/UserContext.tsx
@@ -270,7 +270,6 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
         isClosable: true,
         position: 'top-right',
       });
-      console.log(values);
       getUserActivityTypes();
       getUserContributions();
       getDaoContributions();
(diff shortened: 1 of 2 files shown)


apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -270,6 +270,7 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
         isClosable: true,
         position: 'top-right',
       });
+      console.log(values);
```
jonathanprozzi, line 273: Let's remove this console.log -- good use of it for debugging though!

apps/protocol-frontend/src/utils/validations.ts
```diff
@@ -46,6 +46,7 @@ export const editContributionFormValidation = yup.object({
 export const reportFormValidation = yup.object({
   name: yup.string().required('This field is required.'),
   engagementDate: yup.date(),
+  activityType: yup.string().min(1).required('This field is required.'),
```
jonathanprozzi, line 49: I think we may be able to just use `yup.string().required(`This field is required.')` since we don't need the `.min(1)` for a minimum length

jonathanprozzi (commented):
I'd say let's make those small changes in the comments I added -- we may be able to get this merged in since I can leverage it in another validation issue I'm working on. We can then pair on that issue. You resolved the required part of the validation which closes this Linear ticket!

Nice job!

apps/protocol-frontend/src/utils/validations.ts
jonathanprozzi: did you get a chance to resolve this @NKoech123 ?

NKoech123: Yes sir. Pushed the change a few minutes @jonathanprozzi . Let me know if more edits are required.

alexkeating (approved):

PR #107, opened 2022-07-22
react-table v7 -> v8

amrro:
## Linear Ticket

## Description

## Screencaptures

### DIFF
--- a/apps/protocol-frontend/src/components/AttestationsTable.tsx
+++ b/apps/protocol-frontend/src/components/AttestationsTable.tsx
@@ -156,6 +156,8 @@ const AttestationsTable = ({
     tableHooks
   );
 
+  console.dir(getTableProps());
+
   useEffect(() => {
     setSelectedContributions(selectedFlatRows);
   }, [selectedFlatRows, selectedRowIds]);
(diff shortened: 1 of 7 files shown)


jonathanprozzi:
image

i'm seeing this as the UI right now -- with everything going on this cycle I'm not sure we'd be able to fully work on this and integrate it. 

I agree that v7 isn't ideal but I'll need to take a bit of a deeper look into this since we're losing quite a lot of the current UX here

amrro:
Yeah, it's missing styling from the theme. 
But if you want a rapid comparison between the two, check this: 
https://github.com/chakra-ui/chakra-ui/pull/6339/files

jonathanprozzi:
This is a good starting point though, I can take a look at how we can leverage our existing styles for this since they're in our theme

jonathanprozzi:
Oh I think if you can rebase / merge in the recent staging changes that'll help clean up the diffs too 

Digging into this to see how to approach the styles

jonathanprozzi:
We can handle some of the styles without too much of a headache since they're in our theme already. 

The biggest areas we'd need to handle are replacing/refactoring the search(filter)/sort and the tableHooks that add in the actions (edit/delete and checkboxes for the select)

PR #110, opened 2022-07-25
Build Endpoint to load Attestations in JSON-LD schema

amrro:
## Linear Ticket
- [PRO-276 - Build endpoint to be consumed by Avenue](https://linear.app/govrn/issue/PRO-276/build-endpoint-to-be-consumed-by-avenue)
- [Schemas](https://daostar.org/schemas/)

## Description
Add an express endpoint to load attestations in a json-ld format.

### Response Example: 
```json 
{
   "@context":"http://www.daostar.org/schemas",
   "type":"attestation",
   "contributions":[
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreiaecfm4v7ehebmglltss67mfepy3bgq6pd5kv7ngqv5aone6mukcu",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      },
      {
         "type":"contribution",
         "commit_type":"Pull Request",
         "reference":"ipfs://bafkreieiealvaglltpnxn7qlttgfgj3fxvergqtpz7xkgdys4ybap5hjre",
         "attesterURI":"https://api.govrn.io/apps/profile/0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156"
      }
   ]
}
```


apps/protocol-ld/project.json
```diff
@@ -0,0 +1,51 @@
+{
+  "root": "apps/protocol-ld",
```
alexkeating, line 2: We should document a naming convention, but I am thinking this should probably be called something like voyager-identity-api. Voyager Identity is the name of the metagov working froup

apps/protocol-ld/src/main.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * This is not a production server yet!
+ * This is only a minimal backend to get started.
+ */
```
alexkeating, line 4: This should probably be removed

apps/protocol-ld/src/subgraph.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// const sfsdf = async (contr: any) => {
+//   const detailsUri = ethers.utils.toUtf8String(contr.detailsUri);
+//   const ipfsBlob = await fetchIPFS(detailsUri);
+//   const ipfsText = await ipfsBlob.text();
+//   const contributionDetails = JSON.parse(ipfsText);
+//   return {
+//     detailsUri,
+//   };
+// };
```
alexkeating, line 25: This should probably be removed

apps/protocol-ld/src/subgraph.ts
```diff
@@ -0,0 +1,59 @@
...
+const provider = new ethers.providers.JsonRpcProvider(CHAIN_URL);
+const govrnContract = new GovrnContract(networkConfig, provider);
+
+export const loadContributions: () => Promise<
+  Awaited<LDContribution>[]
+> = async () => {
+  const graphQLClient = new GraphQLClient(SUBGRAPH_ENDPOINT);
+  const client = new GovrnGraphClient(graphQLClient);
+
+  const events = (await client.listAttestations({})).attestations;
```
alexkeating, line 37: We should probably paginate here. A single subgraph call can only fetch 1000 entries

apps/protocol-ld/src/subgraph.ts
```diff
@@ -0,0 +1,59 @@
...
+import { BigNumber, ethers } from 'ethers';
+import { LDContribution } from './types';
+
+const SUBGRAPH_ENDPOINT = process.env.SUBGRAPH_URL;
+const CONTRACT_ADDRESS = process.env.CONTRACT_ADDRESS;
+const CHAIN_URL = process.env.CHAIN_URL;
+
+const networkConfig: NetworkConfig = {
+  address: CONTRACT_ADDRESS,
+  chainId: 2,
```
alexkeating, line 14: I think Gnosis chainId is 100

apps/protocol-ld/src/subgraph.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// load data from subgraph.
+const provider = new ethers.providers.JsonRpcProvider(CHAIN_URL);
+const govrnContract = new GovrnContract(networkConfig, provider);
+
+export const loadContributions: () => Promise<
+  Awaited<LDContribution>[]
+> = async () => {
+  const graphQLClient = new GraphQLClient(SUBGRAPH_ENDPOINT);
+  const client = new GovrnGraphClient(graphQLClient);
```
alexkeating, line 35: This devx is kind of bad. We should probably update the subgraph client, so instead of taking in a GraphQLClient it takes in a chainId. Within the subgraph client we have a constant with a mapping of chainId to subgraph url  that we use to instantiate a GraphQLClient under the hood.

apps/protocol-ld/src/main.ts
```diff
@@ -0,0 +1,28 @@
...
+  res.header('Content-Type', 'application/json');
+  next();
+});
+
+app.get('/api', async (req, res) => {
+  const event = await loadContributions();
+  // json > json-ld
+  const transformed = await transform(event);
+  res.send(JSON.stringify(transformed));
+});
```
alexkeating, line 22: Could the use and get be combined?

apps/protocol-ld/src/subgraph.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // load contributions for data like type..etc.
+  const contrs = await Promise.all(
+    events.map(async (e) => {
+      const contr = await govrnContract.contributions({
+        tokenId: e.contribution.id,
+      });
+      return { ...contr, ...e };
+    })
+  );
```
alexkeating, line 47: Wouldn't it be more efficient to use the subgraph instead of calling the contract for each contribution?

apps/protocol-ld/src/subgraph.ts
```diff
@@ -0,0 +1,59 @@
...
+  return await Promise.all(
+    contrs.map(async (c) => {
+      return {
+        id: Number(BigNumber.from(c.contribution.id.toString()).toString()), // TODO: a better evaluation.
+        attestor: c.attestor,
+        owner: c.owner,
+        detailsUri: ethers.utils.toUtf8String(c.detailsUri),
+      };
+    })
+  );
```
alexkeating, line 58: Does this need to be wrapped in a promise? Nothing is being awaited

apps/protocol-ld/src/subgraph.ts
amrro: Contributions Events from sugraph doesn't include `detailsUri`; `detailsUri` is required for the `reference` field in json-ld response.

apps/voyager-identity-api/src/subgraph.ts
```diff
@@ -0,0 +1,57 @@
...
+  limit?: number;
+}): Promise<LDContribution[]> => {
+  // Load attestations events from subgraph.
+  const page = options.page || 0; // || for NaN value.
+  const limit = options.limit || LIMIT;
+
+  const events = (
+    await client.listAttestations({
+      skip: page * limit,
+      first: limit,
```
alexkeating, line 35: This is better but not the optimal way to do pagination with the graph for large collections. 

https://thegraph.com/docs/en/querying/graphql-api/#example-using-and-2

amrro: There is one issue with using `id_gt` to avoid large `skip`s: `Attest` Event's id is not numeric value, 
https://github.com/Govrn-HQ/govrn-monorepo/blob/b2ca0c2b1f2656770aa694cfcb802e4e9e7ef0c8/apps/govrn-contract-subgraph/src/mapping.ts#L10-L14 

Consequently, `id_gt` clause fails to fetch any entities with values other than 0.

Entity Example: 
```json
{
        "id": "���ԩ�^\u001f��\u000e~(���\u000e0\u0011V-1",
        "attestor": "0xeac5f0d4a9a45e1f9fdd0e7e2882e9f60e301156",
        "confidence": 0,
        "contribution": {
          "id": "0x1"
        }
}
```

alexkeating: We can still do id pagination on non-numeric fields. We can order by id and then paginate using the non-numeric id. We should also update the subgraph to use a correctly encoded address as the attestor in the attestation string which we can capture in a second ticket.

alexkeating: We will create a ticket for this

PR #111, opened 2022-07-25
feature/component-library-docs-pro-218

jonathanprozzi:
## Linear Ticket

- https://linear.app/govrn/issue/PRO-218/component-library-docs
- https://linear.app/govrn/issue/PRO-312/add-and-configure-storybook-addons

## Description

This adds the initial README for the component library, adds stories for an Atom (ErrorMessage) and a Molecule (Input), adds some addons (PRO-312), and fixes the build. 

There are a few other optimizations we can make, but I don't want this to block additional progress to get caught up on Stories. Here are some thoughts for optimizations:

- Handle the Modal and Form context at the Storybook wrapper`preview.js` 
  - This may be necessary before we work on the Modal. The Form should be alright as it is.


jonathanprozzi:
I'm looking into the build issues on this one -- I'm seeing the same issues locally for all `nx` commands I try to run, so going to see what's up with it.

libs/protocol-ui/README.md
```diff
@@ -1,7 +1,80 @@
-# protocol-ui
+# Protocol UI
+
+The `protocol-ui` library is our component library and custom theme that customizes and extends [Chakra UI](https://chakra-ui.com/). The initial focus of this is on the functional components we used to build v such as form elements, modal wrappers, and the custom theme that overrides Chakra's base component styles.
```
alexkeating, line 3: I think v should be v1?

alexkeating (approved):
One small comment

PR #112, opened 2022-07-25
Add Typing to Protocol Frontend

amrro:
## Linear Ticket

[PRO-305 - Add Types for Contribution and Attestation](https://linear.app/govrn/issue/PRO-305/add-types-for-contribution-and-attestation)


apps/protocol-frontend/src/components/ContributionTypesTable.tsx
```diff
@@ -1,52 +1,54 @@
...
             contributionType.activity_type.name
         ).length,
-        lastOccurrence: contributionType.engagementDate,
+        lastOccurrence: contributionType.date_of_engagement,
         activityType: contributionType.activity_type.name,
-        guilds: contributionType.guild,
-        submissionDate: format(
-          new Date(contributionType.date_of_submission),
-          'P'
-        ),
```
alexkeating, line 45: What is the difference between using this formatting and consuming the raw string

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -32,14 +39,18 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
   const govrn = new GovrnProtocol(protocolUrl, { credentials: 'include' });
   const { setModals } = useOverlay();
 
-  const [currentChain, setCurrentChain] = useState<string | null | undefined>(
-    undefined
-  );
+  // const [currentChain, setCurrentChain] = useState<string | null | undefined>(
+  //   undefined
+  // );
```
alexkeating, line 44: This should be removed

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -587,104 +613,134 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
...
+  updateContribution: any;
+  updateLinearEmail: any;
+  updateProfile: any;
+  userActivityTypes: any;
+  userAddress: any;
+  userAttestations: any;
+  userContributions: ContributionItem[];
+  userData: UserData;
+  userDataByAddress: any;
+};
```
alexkeating, line 680: This is a good start, but we should narrow if possible

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -587,104 +613,134 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
...
-        isAuthenticated,
-        isAuthenticating,
-        authenticateAddress,
+        updateProfile,
+        userActivityTypes,
+        userAddress,
+        userAttestations,
+        userContributions,
+        userData,
+        userDataByAddress,
```
alexkeating, line 643: Are these changes made by your editor? If so do you know what settings order this?

libs/protocol-client/src/lib/client/user.ts
```diff
@@ -16,14 +17,14 @@ export class User extends BaseClient {
...
+  public async get(id: number): Promise<UserData> {
+    const user = await this.sdk.getUserCustom({ id });
+    return user.result;
   }
 
   public async update(data: UserUpdateInput, where: UserWhereUniqueInput) {
-    const jobRun = await this.sdk.updateUser({ where, data });
-    return jobRun.updateUser;
+    const user = await this.sdk.updateUser({ where, data });
+    return user.updateUser;
```
alexkeating, line 27: Good rename

libs/protocol-client/src/lib/client/types.ts
```diff
@@ -0,0 +1,50 @@
+export type ContributionItem = {
```
alexkeating, line 1: I don't think we need this object. For something like contribution list the return type is the the result key of type `ListContributionsQuery`. The way we can type this in the protocol frontend is by using ListContributionsQuery['result']. Even if there is no explicit type in protocol types then we should be able to access the underlying key type we return

apps/protocol-frontend/src/contexts/UserContext.tsx
amrro: I made the changes intentionally. It was eye straining to locate some variable, so, I sorted lines them alphabetically.

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -27,39 +26,45 @@ import { useOverlay } from '../contexts/OverlayContext';
...
 import AddAttestationForm from './AddAttestationForm';
+import { ContributionItem } from '@govrn/protocol-client';
 
 const AttestationsTable = ({
   contributionsData,
   setSelectedContributions,
-}: any) => {
+}: {
+  contributionsData: ContributionItem[];
+  setSelectedContributions: (contrs: any[]) => void;
```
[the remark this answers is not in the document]
amrro: This is one of issues with `react-table` v7. The selected contributions are not the type of contriubtion itself. But It's selected rows, using the `Row<D>` type from `react-table`, and `D` is another type composed of the table columns. 

Besides, the `data` mapping from the Contribution Type that is different and simpler. `data` memo includes all the states related to the component, so dividing it into smaller states is inevitable in order to make table more composobale and depend more on the `ContributionItem` type.   

So, You will have to create another type. half information provided is better than nothing. 
We probably should wait for `react-table` migration for `v8` to have clearer refactor.

apps/protocol-frontend/src/components/ContributionTypesTable.tsx
amrro: database format: `2022-07-01T04:00:00.000Z`. 
After formatting: `07/19/2022`

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -27,39 +26,45 @@ import { useOverlay } from '../contexts/OverlayContext';
...
+  // TODO: Divide this into different states.
   const data = useMemo(
     () =>
-      unattestedContributions.map((contribution: any) => ({
+      unattestedContributions.map((contribution) => ({
         id: contribution.id,
-        submissionDate: format(new Date(contribution.date_of_submission), 'P'),
-        engagementDate: format(new Date(contribution.date_of_engagement), 'P'),
+        date_of_submission: contribution.date_of_submission,
+        date_of_engagement: contribution.date_of_submission,
```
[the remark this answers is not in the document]
amrro: This is from the generated types.

apps/protocol-frontend/src/contexts/UserContext.tsx
amrro: This is strenuous task, I think, is going to take a lot of time. 
I am suggesting to gradually update `UserContextType`, while adding and supporting more types

amrro: Also, It seems that User Context provider can't show errors for missing field in `UserContextType`; it continues to work. Not much help.

alexkeating: I think a lot of these types already exist in our protocol-types. Let's take a look at this in a pairing session. If it is a lot of work we can make a new ticket to cover this

libs/protocol-client/src/lib/client/types.ts
```diff
@@ -0,0 +1,5 @@
+import { GetUserCustomQuery, ListContributionsQuery } from '../protocol-types';
+
+export type ContributionItem = ListContributionsQuery['result'][number];
+
+export type UserData = GetUserCustomQuery['result'];
```
alexkeating, line 5: It might be good to adopt a standard naming convention. Like ListContributionsQueryResult instead of contributionItem.

libs/protocol-client/src/lib/client/types.ts
```diff
@@ -0,0 +1,5 @@
+import { GetUserCustomQuery, ListContributionsQuery } from '../protocol-types';
+
+export type ContributionItem = ListContributionsQuery['result'][number];
```
alexkeating, line 3: What is the number meant to do here

amrro: `ListContributionsQuery['result']` returns an array, and we are interested in the element type of that array. 

`T[number]` refer to the element type of the array as `T`.

alexkeating: Good to know

alexkeating: We will move these type to a separate library called ui-types. Which will hold aliases for client types that will be used in frontend libraries. @jonathanprozzi Does prefixing alias types with UI e.g. UIContribution work for you? The main benefit is that our editors can easily find the type, and devs will have a better dev experience (better readibility, discoverability, etc..).

jonathanprozzi: That sounds good to me! We have a types folder in our frontend and I put a “forms.ts” that has all the form related types in it. We can then prefix those as well @alexkeating @amrro

jonathanprozzi: I have these in the other PR. I can update these. So for example we’d do something like UIContributionFormValues instead of ContributionFormValues since those are UI types

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -567,7 +575,6 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
       },
       linear_id: userData.id.toString(), // linear_id exists outside of our db
       name: userData.name,
-      url: userData?.url ?? '',
```
amrro: @alexkeating 
Would you check that? I couldn't find `url` on user.

apps/protocol-frontend/src/components/ReportForm.tsx
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
+// TODO: JP, would you disable this ^ and check errors.
```
amrro, line 2: @jonathanprozzi I can use your help here. Would you disable this flag and check errors? 

They are interesting ones.

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -77,6 +77,7 @@ const AttestationsTable = ({
       {
         Header: 'Status',
         accessor: 'status',
+        // @ts-ignore
```
alexkeating, line 80: If we are adding a bunch of ts-ignores we aren't really solving our typing issues

amrro: This is actually an issue with react-table v7. It has weak typing. And couldn't be fixed.

PR #115, opened 2022-07-26
Change DateTime scalar type from any to string

amrro:
## Linear Ticket
related: 
- [PRO-305 - Add Types for Contribution and Attestation](https://linear.app/govrn/issue/PRO-305/add-types-for-contribution-and-attestation), 
- #112

## Description

`Datetime` scalar type was mapped to type `any`. It's updated to have the type `string | Date`.

### DIFF
--- a/libs/protocol-client/codegen.yaml
+++ b/libs/protocol-client/codegen.yaml
@@ -7,6 +7,8 @@ generates:
       dedupeFragments: true
       nonOptionalTypename: false
       skipTypename: true
+      scalars:
+        DateTime: string | Date
     plugins:
       - 'typescript'
       - 'typescript-operations'
--- a/libs/protocol-client/src/lib/protocol-types.ts
+++ b/libs/protocol-client/src/lib/protocol-types.ts
@@ -13,8 +13,7 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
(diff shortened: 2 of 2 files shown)


jonathanprozzi (commented):
Nice addition of `DateTime`!
For my own clarification -- did the addition of that new DateTime scalar type to the `codegen.yaml` then regenerate the `protocol-types`?

amrro:
> For my own clarification -- did the addition of that new DateTime scalar type to the `codegen.yaml` then regenerate the `protocol-types`?

Yes. Exactly.

amrro:
With these changes, all input types date fields will have the type of `string | date`. 

https://github.com/Govrn-HQ/govrn-monorepo/blob/4e3217b486547e37510e16a934834886e381c873/apps/contract-sync-job/src/main.ts#L100-L108

If you want to provide date as `string`, you will need to use [`Date.prototype.toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) which uses the same ISO formatting used by [prisma](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#datetime): ISO 8601

-----
This is working nicely with creating dates too. Date's constructor accepts `number | string | Date`, which new `DateTime` scalar new JS type, `string | Date`, belongs to.

alexkeating (approved):

PR #116, opened 2022-07-27
feature/bulk-attribute-daos (PRO-236)

jonathanprozzi:
## Linear Ticket

- https://linear.app/govrn/issue/PRO-236/bulk-attribute-to-dao-on-contributions-page

## Description

This adds the feature to attribute multiple Contributions to a DAO at once. This is done via the checkboxes that we have for minting/attesting -- the user then selects an _Attribute to DAO_ button instead of _Mint_ which then pops up a modal for them to choose. This uses our existing UX pattern with the modal popping up for confirming/entering details on a bulk action.

I was talking with @scottrepreneur that a pattern we may want to employ as an enhancement is to add the dropdown DAO select box to the controls (where the buttons are on the table) and then allow a user to do this directly from the Contributions table itself (only on Desktop). 

Now that this feature is in functionally we can explore those types of enhancements.

I had to do some work to get the data to persist and populate when coming from React Table. This will definitely be improved with v8 like @amrro has been exploring. v7 makes it a bit tricky to get the original data out and it looks like v8 will improve that for sure.

## Screencaptures

https://user-images.githubusercontent.com/9438776/181275258-9222b211-356a-4be2-ab01-9c2ae22f250e.mp4


jonathanprozzi:
Going to add typing to this PR -- looking to minimize conflicts with the open types PR but also want to add some easy wins into this one @alexkeating

apps/protocol-frontend/src/components/BulkDaoAttributeModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Select } from '@govrn/protocol-ui';
+import { Stack, Button, Text, Progress } from '@chakra-ui/react';
+import { useForm, SubmitHandler } from 'react-hook-form';
+import { yupResolver } from '@hookform/resolvers/yup';
+import { useUser } from '../contexts/UserContext';
+import { editContributionFormValidation } from '../utils/validations';
+import { BulkDaoAttributeFormValues } from '../types/forms';
+
+interface BulkDaoAttributeModalProps {
+  contributions: any;
```
alexkeating, line 11: Would merging in Amrros pr resolve these

jonathanprozzi: Yup, I think that if we merge in Amrro's PR first we should be able to then fill in all the Contribution, Contribution[], Attestation, and Attestation[] types

If we do that first, then I think we'll be in better shape with types except for the ones stemming from React Table v7

PR #117, opened 2022-07-27
Atom stories

NKoech123:
## Linear Ticket
PRO-303

https://linear.app/govrn/issue/PRO-303/stories-for-the-form-atoms
## Description

Added stories on FormLabel and HelperText. Got issues with HelperText. @jonathanprozzi 

## Screencaptures
Screen Shot 2022-07-27 at 11 08 15 AM
Screen Shot 2022-07-27 at 11 08 35 AM

### DIFF
--- a/libs/protocol-ui/src/stories/atoms/FormContainer.stories.tsx
+++ b/libs/protocol-ui/src/stories/atoms/FormContainer.stories.tsx
@@ -1,1 +1,0 @@
-// TODO: need to create this component before creating a story
--- a/libs/protocol-ui/src/stories/atoms/FormContainer.stories.tsx
+++ b/libs/protocol-ui/src/stories/atoms/FormContainer.stories.tsx
@@ -1,0 +1,1 @@
+// TODO: need to create this component before creating a story
--- a/libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
+++ b/libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
@@ -1,0 +1,25 @@
+import { Story } from '@storybook/react';
+import React from 'react';
+import {FormLabel, FormLabelProps} from '../../'
+
+export default {
+    title: 'Components/Atoms/FormLabel',
+    component: FormLabel,
+    argTypes: {
+        label:{
+            control: { type: 'text' },
+        },
+        //color: {control:'color'},
+    },
+};
+
+const Template: Story<FormLabelProps> = (args) => <FormLabel {...args} />;
+    
+const Default = Template.bind({})
+Default.args = {
+    label: 'DAO',
+    //color: 'gray.800'
+
+};
+
+export { Default };
--- a/libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
+++ b/libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
@@ -1,25 +1,24 @@
 import { Story } from '@storybook/react';
 import React from 'react';
-import {FormLabel, FormLabelProps} from '../../'
+import { FormLabel, FormLabelProps } from '../../';
 
 export default {
-    title: 'Components/Atoms/FormLabel',
-    component: FormLabel,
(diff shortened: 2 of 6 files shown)


jonathanprozzi:
@NKoech123 I've looked into this a bit and I'm not sure why we're seeing it only on this Story. It looks like it's an issue with the Chakra Storybook plugin but we use that throughout the Storybook. I'm going to keep looking

libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+import {FormLabel, FormLabelProps} from '../../'
+
+export default {
+    title: 'Components/Atoms/FormLabel',
+    component: FormLabel,
+    argTypes: {
+        label:{
+            control: { type: 'text' },
+        },
+        //color: {control:'color'},
```
jonathanprozzi, line 12: We can add this back in but this will add a color picker that'll let folks see what this looks like with _any_ color available. We may want to lock it down to only use Chakra/our color tokens, but should be ok to include this for now

libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    },
+};
+
+const Template: Story<FormLabelProps> = (args) => <FormLabel {...args} />;
+    
+const Default = Template.bind({})
+Default.args = {
+    label: 'DAO',
+    //color: 'gray.800'
```
jonathanprozzi, line 21: Can add this back in so it'll use the `gray.800` default value

libs/protocol-ui/src/stories/atoms/FormContainer.stories.tsx
```diff
@@ -0,0 +1 @@
+// TODO: need to create this component before creating a story
```
jonathanprozzi, line 1: Let's delete this Story until we have the actual component

jonathanprozzi (commented):
Looks good! A few comments to resolve, but we shouldn't merge until we figure out the Chakra issue on the HelperText.

I'm still looking into it -- we'll merge this in as soon as that's resolved

jonathanprozzi:
@NKoech123 fixed this -- was a really strange issue with how Chakra handles isolated form elements.

Basically, the issue was that for _some_ form elements such as the _HelperText_ it inherits a style object from the parent, so it needed to be wrapped in a <FormControl/>

I fixed this directly on the PR

NKoech123:
That really makes sense. I just checked Chakra documentation and saw _HelperText_  is supposed to inherit from _FormControl_. Thanks again for catching that and fixing it!

libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
NKoech123: I see you uncommented that line 12. Thanks!

libs/protocol-ui/src/stories/atoms/FormLabel.stories.tsx
NKoech123: I see you uncommented that line 21. Thanks!

libs/protocol-ui/src/stories/atoms/FormContainer.stories.tsx
NKoech123: I just deleted this. Resolved.

NKoech123:
## Resolved: Summary
- chakra-UI component Inheritance issue resolved(thanks @jonathanprozzi )
- deleted file 'FormContainer.stories.tsx'

## ScreenCaptures

Screen Shot 2022-07-30 at 12 51 32 PM
Screen Shot 2022-07-30 at 12 51 43 PM
Screen Shot 2022-07-30 at 12 51 55 PM

jonathanprozzi (commented):
Good to go!

PR #123, opened 2022-07-31
Custom resolver for contribution count

Flip-Liquid:
## Linear Ticket

[Pro-293](https://linear.app/govrn/issue/PRO-293/preprocess-data-for-mvp-contributor-dashboard)

## Description

- Add a custom resolver to preprocess a query to get the number of contributions for a user in a given timeframe
- Update client library readme to reflect that the client needs regeneration if resolvers are edited
- Update permissions for retrieving reporting channel field

## Screencaptures

### DIFF
--- a/apps/protocol-api/src/main.ts
+++ b/apps/protocol-api/src/main.ts
@@ -31,7 +31,7 @@ const typeSchema = buildSchemaSync({
 
 // TODO: Figure out is there is a way to genralize this
 //
-const OwnsData = rule()(async (parent, args, ctx, info) => {
+const ownsData = rule()(async (parent, args, ctx, info) => {
   return true;
 });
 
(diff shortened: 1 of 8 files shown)


apps/protocol-api/src/prisma/resolvers/User.ts
```diff
@@ -127,4 +148,20 @@ export class UserCustomResolver {
       where: { address: { equals: address } },
     });
   }
+
+  @TypeGraphQL.Query((_returns) => Int, { nullable: false })
+  async getContributionCountForUser(
+    @TypeGraphQL.Ctx() ctx: Context,
+    @TypeGraphQL.Args() args: GetUserContributionCountArgs
+  ) {
+    return await getPrismaFromContext(ctx).contribution.count({
```
alexkeating, line 157: I think a group would work better here. This, as far as I can tell, provides a single count, but I think the client is going to want to get the counts for multiple days or weeks in a single call because I believe we will use these to populate a contribution calendar and bar chart.


Example taken from .prisma/client/index.d.ts
```
    /**
     * Group by Contribution.
     * Note, that providing `undefined` is treated as the value not being there.
     * Read more here: https://pris.ly/d/null-undefined
     * @param {ContributionGroupByArgs} args - Group by arguments.
     * @example
     * // Group by city, order by createdAt, get count
     * const result = await prisma.user.groupBy({
     *   by: ['city', 'createdAt'],
     *   orderBy: {
     *     createdAt: true
     *   },
     *   _count: {
     *     _all: true
     *   },
     * })
     * 
    **/

```

apps/protocol-api/src/main.ts
```diff
@@ -77,15 +78,15 @@ const permissions = shield(
...
+      createUserCustom: or(hasToken, and(ownsData, isAuthenticated)),
       createUserOnChainAttestation: isAuthenticated,
-      deleteContribution: or(OwnsData, isAuthenticated),
+      deleteContribution: or(ownsData, isAuthenticated),
       updateGuild: hasToken,
       updateUser: hasToken,
-      updateUserContribution: and(OwnsData, isAuthenticated),
-      updateUserCustom: and(OwnsData, isAuthenticated),
+      updateUserContribution: and(ownsData, isAuthenticated),
+      updateUserCustom: and(ownsData, isAuthenticated),
```
alexkeating, line 89: Great tech debt cleanup!

alexkeating (commented):
Looks good, but I think we may need to use a different method from prisma

apps/protocol-api/src/prisma/resolvers/User.ts
alexkeating: https://github.com/prisma/prisma/discussions/11692

alexkeating: Probably has to be done with a raw sql query

PR #125, opened 2022-08-01
Add Transaction Hash to the Contribution Databse Model

amrro:
## Linear Ticket

[PRO-327 - Add transaction hash to the database](https://linear.app/govrn/issue/PRO-327/add-transaction-hash-to-the-database)

## Description

For minted contribution, `tx_hash` added to the database model of `Contribution`.


libs/protocol-client/src/lib/protocol-types.ts
```diff
@@ -2188,6 +2187,7 @@ export type Contribution = {
   status: ContributionStatus;
   status_id: Scalars['Int'];
   tweet?: Maybe<TwitterTweet>;
+  tx_hash?: Maybe<Scalars['String']>;
```
alexkeating, line 2190: I think the only thing missing is to make sure this field is added to the permissions in the protocol-api.

PR #142, opened 2022-08-06
feature/frontend-add-cypress (PRO-295)

jonathanprozzi:
## Linear Ticket

- https://linear.app/govrn/issue/PRO-295/add-end-to-end-testing

## Description

- This adds Cypress e2e testing to the `protocol-frontend` via the `protocol-frontend-e2e` package.
- Installing the `@nrwl/cypress` and `@nrwl/jest` 13.8.4 to match our `nx` version -- these both had to match or we'd have configuration issues. We can then migrate all of our nx plugin packages at once
- Had to change a few config files out of the box to support Cypress v10 -- the entry point for the config is `cypress.config.ts` and this has to match what's run in the `project.json` for the `cypressConfig` 
- Running `nx run protocol-frontend-e2e:e2e` runs Cypress in the command line, whereas `nx run protocol-frontend-e2e:e2e --watch` opens the Cypress UI if you have it installed (https://www.cypress.io/) -- this can run in Chrome or Electron
- I changed out `HomeShell` to include an h1 that is visually hidden. This was originally in a header that is now unused, so our first test caught an error!

## Screencaptures

https://user-images.githubusercontent.com/9438776/183259911-99a7d0e2-b266-4058-8a8a-5c44eeaa2b1d.mp4

Initial test passes. We now can add more!


jonathanprozzi:
The `yarn.lock` conflict is likely due to me changing the `@nrwl/jest` version -- @alexkeating

PR #153, opened 2022-08-11
Fix Avenue Endpoint 

amrro:
## Linear Ticket
[PRO-372 - Avenue endpoint returns no contributions](https://linear.app/govrn/issue/PRO-372/avenue-endpoint-returns-no-contributions)

## Description
There are contributions on the subgraph but the endpoint returns 0 contributions.
### Fixed
- `page` wasn't used and passed at null to the client. 
- `skip` limit: The `skip` argument must be between 0 and 5000.

### DIFF
--- a/apps/voyager-identity-api/src/subgraph.ts
+++ b/apps/voyager-identity-api/src/subgraph.ts
@@ -23,10 +23,15 @@ const graphQLClient = new GraphQLClient(SUBGRAPH_ENDPOINT);
 const client = new GovrnGraphClient(graphQLClient);
 
 export const loadContributions = async (options: {
+  address?: string;
   page?: number;
   limit?: number;
 }): Promise<LDContribution[]> => {
-  // Load attestations events from subgraph.
+  if (!options.address) {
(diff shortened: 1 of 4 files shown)


apps/voyager-identity-api/src/subgraph.ts
```diff
@@ -26,12 +27,15 @@ export const loadContributions = async (options: {
   limit?: number;
 }): Promise<LDContribution[]> => {
   // Load attestations events from subgraph.
-  const page = options.page || 0; // || for NaN value.
-  const limit = options.limit || LIMIT;
+  const page = Math.abs(options.page || 1); // || for NaN value.
+  const limit = Math.abs(options.limit || LIMIT);
+  const skip = Math.min((page - 1) * limit, SKIP_LIMIT);
```
alexkeating, line 32: This works for now, but we need to make this pagination more robust. And instead of taking the minimum value we should throw an error. Being explicit is better because then the client can report why their query is failing rather than getting the same values back.

PR #156, opened 2022-08-12
Fix issue

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-385/look-into-usercontext-issues-with-the-user-id

## Description

Fix long load times, because user is not set in query


jonathanprozzi (approved):
great catch

jonathanprozzi (commented):
I think we'll actually need this on the 
```

PR #161, opened 2022-08-16
Show Spinner for Loading Initial User Data at Start Page

amrro:
## Linear Ticket

[PRO-318 - Handle loading state](https://linear.app/govrn/issue/PRO-318/handle-loading-state)

## Description
For the time being taken to load the user's initial data, the start page show different different states that is not accurate till the data is loaded; A loading indicator added waiting for user's data to finish loading.


jonathanprozzi:
Great addition. I think if we end up using Spinners enough (we are) we can also add one to the component library

jonathanprozzi (approved):
🎉 big UX gain from this

PR #164, opened 2022-08-16
Update Dependencies

amrro:
## Linear Ticket

[PRO-436 - Migrate Nx to latest version](https://linear.app/govrn/issue/PRO-436/migrate-nx-to-latest-version) 

## Description
Update all nx plugins besides dependencies.
updating react dependencies is excluded.


apps/protocol-frontend/src/components/ContributionDetailShell.tsx
```diff
@@ -76,7 +76,7 @@ const ContributionDetailShell = ({
                   : `${contribution?.status.name} 👀`}
               </Badge>
               <Text fontSize="sm" color="gray.600">
-                {contribution?.date_of_engagement}
+                {`${contribution?.date_of_engagement}`}
```
alexkeating, line 79: Why do we need this change?

apps/protocol-frontend/src/components/CreateWaitlistUserForm.tsx
```diff
@@ -12,7 +12,7 @@ import { CreateUserFormValues } from '../types/forms';
 
 const useYupValidationResolver = (userValidationSchema: any) =>
   useCallback(
-    async data => {
+    async (data: any) => {
```
alexkeating, line 15: Is it possible to use a more specific type?

libs/protocol-client/src/lib/client/mint-event-example.json
```diff
@@ -0,0 +1,40 @@
+{
+  "eventFragment": {
```
alexkeating, line 2: Is this being used anywhere?

apps/protocol-frontend/src/components/ContributionDetailShell.tsx
amrro: It's throwing the following error: 
`TS2322: Type 'string | Date' is not assignable to type 'ReactNode'`

PR #165, opened 2022-08-16
feature/user-dashboard-heat-map-v1

jonathanprozzi:
## Linear Ticket

- [PRO-375](https://linear.app/govrn/issue/PRO-375/create-user-dashboard-page-shell)
- [PRO-376](https://linear.app/govrn/issue/PRO-376/create-a-controlled-selectmulti-select-component-for-the-ui-library)
- [PRO-356](https://linear.app/govrn/issue/PRO-356/add-user-heat-map)
- [PRO-359](https://linear.app/govrn/issue/PRO-359/add-user-dashboard-filters)

## Description

This is the first PR for the User Dashboards, and adds the following:
- The `feature/dashboard` route (will change upon release)
- User Dashboard shell, Filter, and Heatmap
- `getUserContributionsCount` to the `UserContext`
- `nivo` charts package for the chart

Notes: 
- Mobile is still a work in progress
- Filters are only applying 1 at a time using the first index of the array -- Have to update the resolver to support an array of ids. Default is to show all Contributions across all DAOs
- I added support for a flexible date range on the calendar chart. It displays 1 year prior to the current date similar to GitHub, but I added ability to select a custom date range (removed this from the UI, but we can add it back in)

I wanted to get this PR up -- I know I still need to fix the filter issue but didn't want to further delay this PR -- @alexkeating I did some experiments with that last night and didn't have much luck with the raw SQL query that this query is using

## Screencaptures

https://user-images.githubusercontent.com/9438776/184909035-a8610531-e7df-427c-9e22-4d4f4e2da080.mp4


alexkeating:
Beefy

apps/protocol-frontend/src/components/DashboardShell.tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// fetching will need to happen on this page, will refetch after each filter update
+// passing in the User so that the name is available upon initial render
+
+const DashboardShell = ({ user }: DashboardShellProps) => {
+  const { allDaos, getUserContributionsCount } = useUser();
+  const [contributionsCount, setContributionsCount] = useState([]);
+  const [dateRange, setDateRange] = useState<number>(52);
+  const [selectedDaos, setSelectedDaos] = useState<any>([]);
```
alexkeating, line 22: We should remove this any as it looks like it is a Set of {value, label}

apps/protocol-frontend/src/components/DashboardShell.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  useEffect(() => {
+    const fetchHeatMapCount = async () => {
+      const contributionsCountResponse = await getUserContributionsCount({
+        startDate: subWeeks(new Date(), dateRange),
+        endDate: new Date(),
+        guildIds: selectedDaos[0].value === null ? null : selectedDaos[0].value,
```
alexkeating, line 29: we should probably check if selected daos has a length 0

apps/protocol-frontend/src/components/DashboardShell.tsx
jonathanprozzi: Good call!

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -65,16 +70,17 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
     [],
   );
   const [allDaos, setAllDaos] = useState<UIGuild[]>([]);
+  const [userContributionsDateRangeCount, setUserContributionsDateRangeCount] =
+    useState<any>(null);
```
alexkeating, line 74: The any should also be removed here. We can add a ui type and then this will be the ui type and null

libs/protocol-ui/src/components/molecules/ControlledSelect/ControlledSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+  defaultValue,
+  options,
+  isMulti,
+  isClearable,
+  onChange,
+  isDisabled,
+  value,
+}: ControlledSelectProps) => {
+  // const theme = useTheme();
+  // const formThemeColors = customSelectThemeColors(theme);
```
alexkeating, line 68: Staggled comments

libs/protocol-ui/src/components/molecules/ControlledSelect/ControlledSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+          isMulti={isMulti}
+          onChange={onChange}
+          isDisabled={isDisabled}
+          // theme={localTheme => ({
+          //   ...localTheme,
+          //   colors: {
+          //     ...formThemeColors,
+          //   },
+          // })}
+          // styles={customSelectStyles}
```
alexkeating, line 97: Probably should be removed

jonathanprozzi: Good call -- can add these back in when I'm ready to use the theme

PR #166, opened 2022-08-16
polyfill buffer for WalletConnect on RainbowKit

scottrepreneur:
## Linear Ticket

Fixes [PRO-388](https://linear.app/govrn/issue/PRO-388/walletconnect-connection-on-rainbowkit-isnt-working), Looks like we need to polyfill this for Vite.

## Description

RainbowKit [requires some polyfills for Vite](https://www.rainbowkit.com/docs/installation#additional-build-tooling-setup).

## Status

- [x] walletconnect working with polyfill


scottrepreneur:
Not sure why the action is failing where it is however.

jonathanprozzi:
> Not sure why the action is failing where it is however.

This is a strange spot to fail. Going to look into it shortly -- did it build fine locally?

scottrepreneur:
build works for me locally

scottrepreneur:
There was a `govrn-monorepo` nested submodule?? Seems weird, but works w/o

package.json
```diff
@@ -15,6 +15,7 @@
     "@chakra-ui/react": "^1.8.8",
     "@emotion/react": "^11.9.3",
     "@emotion/styled": "^11.9.3",
+    "@esbuild-plugins/node-globals-polyfill": "^0.1.1",
```
amrro, line 18: I would like to grab your attention that this lib is not well maintained and have some issues. 
You can find more here 
- https://github.com/prettier/prettier/issues/12054
- https://github.com/rollup/rollup/issues/2881


https://github.com/remorses/esbuild-plugins/blob/63b1ee1f47c7b0deb46bd71ec13d4a538e443d4d/node-modules-polyfill/src/index.ts#L39-L44


Simply put: it just paste couple of imports to every file which can lead to complex implications. 

CC: @alexkeating @jonathanprozzi

scottrepreneur: Are you seeing a viable solution within these? Should we log this as debt and revisit?

amrro: Actually I don't have a viable solution to suggest.
We need Keating & JP's input on that.

jonathanprozzi (approved):

apps/protocol-frontend/src/polyfills.ts
```diff
@@ -1,7 +1,7 @@
-/**
- * Polyfill stable language features. These imports will be optimized by `@babel/preset-env`.
- *
- * See: https://github.com/zloirock/core-js#babel
- */
-import 'core-js/stable';
-import 'regenerator-runtime/runtime';
```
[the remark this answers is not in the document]
scottrepreneur: Preserving comment in case we run into the issue that required these.

PR #168, opened 2022-08-17
Feature/support multiple guildws

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-386/change-filter-resolver-to-support-more-than-1-dao-for-the-heatmap

## Description

Support multiple guild ids


jonathanprozzi (approved):
Nice! I'll need to fix my heat map PR after this since it'll change the expected types

PR #175, opened 2022-08-22
Fix userData null checks fails in UserContext

amrro:
## Linear Ticket

[PRO-401 - getUserActivityTypes() fails to load](https://linear.app/govrn/issue/PRO-401/at-app-startup-getuseractivitytypes-fails-to-load)

Making `userData`'s type as `{} as UIUser` made all null checks fail. The type is changed to accept `null` value too.

### Example: 
```
Error: getUserContributions has no userData.id
    getUserContributions UserContext.tsx:166
    UserContextProvider UserContext.tsx:714
    unstable_runWithPriority scheduler.development.js:468
    unstable_runWithPriority scheduler.development.js:468
    authFlow AuthContext.tsx:92
    AuthContextProvider AuthContext.tsx:110
    unstable_runWithPriority scheduler.development.js:468
    unstable_runWithPriority scheduler.development.js:468
    <anonymous> main.tsx:30
UserContext.tsx:184:14

```

### DIFF
--- a/apps/protocol-frontend/src/contexts/UserContext.tsx
+++ b/apps/protocol-frontend/src/contexts/UserContext.tsx
@@ -72,7 +72,7 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
   );
   const [isUserLoading, setUserLoading] = useState(false);
 
-  const [userData, setUserData] = useState<UIUser>({} as UIUser);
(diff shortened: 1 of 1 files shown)


jonathanprozzi (approved):
great catch Amrro!

PR #179, opened 2022-08-25
fix/prevent-attestation-not-minted (PRO-374)

jonathanprozzi:
## Linear Ticket

- [PRO-374](https://linear.app/govrn/issue/PRO-374/remove-ability-to-attest-to-non-minted-contributions)

## Description

- Filters out non-minted Contributions so users can't Attest to them
- Small UX tweaks to make sure that there wasn't unexpected behavior -- the _My Attestations_ will remain unfiltered so users will see past Attestations

## Screencaptures

filtered-minted

filtered-minted-2

### DIFF
--- a/apps/protocol-frontend/src/components/GlobalFilter.tsx
+++ b/apps/protocol-frontend/src/components/GlobalFilter.tsx
@@ -37,7 +37,7 @@ const GlobalFilter = ({
         </InputLeftElement>
         <Input
           placeholder={`Search ${rowCount} ${
-            rowCount > 1 ? 'Contributions' : 'Contribution'
+            rowCount === 1 ? 'Contribution' : 'Contributions'
           }`}
           value={value || ''}
           onChange={(e) => {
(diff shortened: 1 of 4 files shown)


apps/protocol-frontend/src/components/GlobalFilter.tsx
```diff
@@ -37,7 +37,7 @@ const GlobalFilter = ({
         </InputLeftElement>
         <Input
           placeholder={`Search ${rowCount} ${
-            rowCount > 1 ? 'Contributions' : 'Contribution'
+            rowCount === 1 ? 'Contribution' : 'Contributions'
```
amrro, line 40: **Not so important**: 
I am starting to notice this frequent plural pattern; I think we can make a function: `plural(countable, count)` or `pluralize(countable, count)` to be more clear & readable.

amrro (approved):
LGTM.

PR #180, opened 2022-08-25
Staging to Master

amrro:
- Add infura subdomain to contract sync job deploy


apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -535,33 +533,24 @@ export class ContributionCustomResolver {
...
+             date(date_of_engagement),
+             count(date_of_engagement)      as count
+      FROM "Contribution" c
+             LEFT JOIN "GuildContribution" as gc
+                       ON c."id" = gc."contribution_id"
+             LEFT JOIN "Guild" as g
+                       ON g."id" = gc."guild_id"
+      WHERE (c."date_of_engagement" BETWEEN ${start} AND ${end}
+        AND c."user_id" = ${user_id}
+        AND (NOT c.id IN (SELECT contribution_id as id FROM "GuildContribution") OR gc."guild_id" in (${Prisma.join(
```
alexkeating, line 550: Why do we need the subquery? Also couldn't the or be a conditional string that is added if there are guild ids? @amrro

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -56,10 +57,17 @@ const AttestationsTable = ({
 }) => {
   const { userData } = useUser();
 
-  const nonUserContributions = _.filter(contributionsData, function (a) {
-    return a.user.id !== userData?.id;
+  const filteredMintedContributions = _.filter(contributionsData, function (a) {
```
alexkeating, line 60: A better name may be mintedContributions. If there isn't another variable named minted contributions

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -693,43 +709,28 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
 
   useEffect(() => {
     if (address && isAuthenticated) {
-      getUserByAddress();
+      getUserByAddress().then();
```
alexkeating, line 712: Why add then here if there is no logic? A non awaited promise should do the same as one with then right?

apps/protocol-api/src/prisma/resolvers/Contribution.ts
amrro: > Why do we need the subquery?

We need to count the contributions that doesn't belong to any guild. So, the subquery loads all `contribution_id`s and counts the ones that does `NOT` belong to it. 

> Also couldn't the or be a conditional string that is added if there are guild ids?

Yes, we can. Is this going to impact performance?

apps/protocol-frontend/src/contexts/UserContext.tsx
amrro: Yes, it should. Stupid mistake! I got confused between `async`/`await` and `then()` or `catch()`, I thought they have the same behavior, forgetting that `then()` or `catch()` takes a function.

In `async/await`, for non-awaited promises it wouldn't handle the rejected case (in case of error) or propagate the error, for better debugging.

[Playground Example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBATgUwFYOFBATACnEBbASwgRgF4YBDCATzGBgAoBKMgPhgG8YAoGGKABa4A7jDAJRAUTi44DAOQAxCgQA2mGDnxEE8pgG5uAX27dQkWADMArqtWW167LkLEylGnUYtS7Lr3gEKGs4MBgAFgAmQxMzcGgYPBUw8ipaemY2Th4+cwTVAksSVOEVK1t7R0wtVwRmQ1z4kHUAOlUQAHMGDgKiowNTKDhqbICAejHKUoJYRBQ0apcdevHJudR0Z21iFb4JwPnNmuWmFsEEMBWjGGAKKGABRgQWf0bIZoQ2zoZnmH2hggdDoIRAYAD8xgCeQ+Xy6ACIAFxIhHrBYYZEIuFMSGmJIES4DIA)

apps/protocol-api/src/prisma/resolvers/Contribution.ts
alexkeating: True, but wouldn't it be more effcient to do an OR gc."guild_id" is NULL?

PR #181, opened 2022-08-25
Display Activity Types That is Created by User or Associated Guilds

amrro:
## Linear Ticket
[PRO-395 - Only show user activity types and activity types related to a dao that users are in.](https://linear.app/govrn/issue/PRO-395/only-show-user-activity-types-and-activity-types-related-to-a-dao-that)

## Description
Show only `ActivityType`s that is: 
- created by the same user, 
- or from associated guilds that this user is in.


jonathanprozzi (approved):

apps/protocol-api/src/prisma/resolvers/ActivityTypes.ts
```diff
@@ -0,0 +1,59 @@
...
+    return await prisma.$queryRaw<ActivityTypesByUser>`
+      SELECT at.*, ua.user_id, gat.guild_id
+      FROM "ActivityType" at
+             LEFT JOIN "UserActivity" ua
+                       ON at.id = ua.activity_type_id
+             LEFT JOIN "GuildActivityType" gat
+                       ON at.id = gat.activity_type_id
+      WHERE ua.user_id = ${user_id}
+         OR (gat.guild_id IN (SELECT guild_id FROM "GuildUser" WHERE user_id = ${user_id}))
+      ORDER BY at.name;`;
```
alexkeating, line 57: This can be replaced by something like 

```

    prisma.activityType.findMany({
      OR: [
        {
          users: {
            some: {
              user_id: { equals: 1 },
            },
          },
        },
        {
          guilds: {
            some: {
              guild_id: { equals: 1 },
            },
          },
        },
      ],
    });
```

apps/protocol-api/src/prisma/resolvers/ActivityTypes.ts
```diff
@@ -0,0 +1,59 @@
...
+  @TypeGraphQL.Field(_type => String)
+  updatedAt: string;
+}
+
+@TypeGraphQL.Resolver(_of => ActivityType)
+export class ActivityTypesCustomResolver {
+  @TypeGraphQL.Query(_returns => [ActivityTypesByUser], {
+    nullable: false,
+  })
+  async getActivityTypesByUser(
```
alexkeating, line 42: We can change this to list and the params should be the same as findMany

apps/protocol-api/src/prisma/resolvers/ActivityTypes.ts
```diff
@@ -0,0 +1,59 @@
...
+    @TypeGraphQL.Ctx() { prisma }: Context,
+    @TypeGraphQL.Args() args: GetActivityTypesPerUserAndDAOsArgs,
+  ) {
+    const user_id = args.where.userId;
+
```
alexkeating, line 47: We should check that the user in the context is the same as the user_id and also verify the user is part of the requested guilds.

libs/protocol-client/src/lib/client/custom.ts
```diff
@@ -42,4 +43,11 @@ export class Custom extends BaseClient {
       await this.sdk.getContributionCountByDateForUserInRange({ where: args });
     return contributionCount.result;
   }
+
+  public async getActivityTypesByUser(args: GetActivityTypesPerUserAndDaOs) {
```
alexkeating, line 47: I don't think this is necessary with the resolver changes

alexkeating:
I think you overwrote your changes. I no longer see any of the resolver changes

alexkeating:
Looks great!

PR #182, opened 2022-08-26
Link ActivityType to Its Creator

amrro:
## Linear Ticket

[PRO-403 - Reporting Contribution doesn't link the ActivityTypes to its Creator/Guild](https://linear.app/govrn/issue/PRO-403/reporting-contribution-doesnt-link-the-activitytypes-to-its)

## Description

When creating a new `ActivityType`, it updates or link this to its creator by updating `UserActivity` Table. 

A new `getOrCreateActivityType` custom mutation has been added to handle creating `ActivityType`s.

### DIFF
--- a/apps/protocol-api/src/prisma/resolvers/ActivityTypes.ts
+++ b/apps/protocol-api/src/prisma/resolvers/ActivityTypes.ts
@@ -1,0 +1,73 @@
+import * as TypeGraphQL from 'type-graphql';
+import { ActivityType } from '../generated/type-graphql';
+import { Context } from './types';
+
+@TypeGraphQL.InputType('GetOrCreateActivityTypeInput', {
+  isAbstract: true,
+})
+export class GetOrCreateActivityTypeInput {
+  @TypeGraphQL.Field(_type => String)
+  activityTypeName: string;
+
+  @TypeGraphQL.Field(_type => Number, { nullable: true })
+  userId?: number;
+}
+
+@TypeGraphQL.ArgsType()
+export class GetOrCreateActivityTypeArgs {
+  @TypeGraphQL.Field(_type => GetOrCreateActivityTypeInput, {
+    nullable: false,
+  })
+  data!: GetOrCreateActivityTypeInput;
+}
+
+@TypeGraphQL.Resolver(_of => ActivityType)
+export class ActivityTypesCustomResolver {
+  @TypeGraphQL.Mutation(_returns => ActivityType, { nullable: false })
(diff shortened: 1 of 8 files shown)


amsos:
quick question: with the pr does it update both the UserActivity and GuildActivityType table? 

I think when a user creates the activity, for now we just want it to link it to the user, not necessarily the guild.  

In previous versions, we've learned that auto updating it for the guild as well creates data explosions.  For example, 10 people all have a different way of writing `taking notes`.  And if we automatically update this as guild activity as well, the guild then has 10 different ways.

Consolidating and moving something from a user activity type to a guild activity type is actually something that the research team is working on.

amrro:
Thanks a lot for your input. It answered a lot of my questions while implementing this. 
This is one of the reasons why I brought the idea of owner/admin/founder of a guild. 

I couldn't find a place that updates the Guild's `ActivityType`. Totally agree with you, this makes a lot of sense.

apps/protocol-api/src/prisma/resolvers/ActivityTypes.ts
```diff
@@ -0,0 +1,73 @@
...
+        },
+        user: {
+          connect: {
+            id: args.data.userId,
+          },
+        },
+      },
+    });
+
+    return await prisma.activityType.findUnique({
```
alexkeating, line 67: Does userActivity type not have an activity type field? We may be able to remove this request.

apps/protocol-api/src/prisma/resolvers/index.ts
```diff
@@ -1,8 +1,10 @@
+import { ActivityTypesCustomResolver } from './ActivityTypes';
 import { AttestationResolver } from './Attestation';
 import { ContributionCustomResolver } from './Contribution';
 import { UserCustomResolver } from './User';
 
 export const customResolvers = [
+  ActivityTypesCustomResolver,
```
alexkeating, line 7: We should rename the resolver to be singular. ActivityTypeCustomResolver

libs/protocol-client/src/lib/client/custom.ts
```diff
@@ -16,6 +16,12 @@ export class Custom extends BaseClient {
   }
 
   public async createUserContribution(args: UserContributionCreateInput) {
+    await this.sdk.getOrCreateActivityType({
```
alexkeating, line 19: Should we get the response of this function call and update the args for createUserContribution?

alexkeating (commented):
Some small comments

PR #191, opened 2022-09-13
feature/add-unassigned-contributions-dashboard (PRO-399)

jonathanprozzi:
## Linear Ticket

[PRO-399](https://linear.app/govrn/issue/PRO-399/add-unassigned-contribution-support-to-dashboard-filters)

## Description

This adds in the 'Unassigned' option to the Dashboard filter multi-select.

On initial load, all contributions show for the user. When they begin adding and removing filters, the contributions will only display for the selected DAOs. If a user includes 'Unassigned' in this selection, the Unassigned contributions will be included.

If the user removes _all_ filters, all contributions will show like in the initial load. This required an additional path -- I added this because I think users will likely want the continuity where the filters are empty at initial load and all contributions load, so   when there are no DAOs in the filter all contributions will load. This causes a slight UI shift that I've reduced, but it's still apparent as it reloads the contributions for the heat map. 

We may want to consider having this state after initial load be empty if a user has no filters added.

## Screencaptures

https://user-images.githubusercontent.com/9438776/189915764-30905fc9-eabb-4686-b472-ee78e857c95d.mp4


jonathanprozzi:
Having some issues uploading my video -- will try again in a bit

apps/protocol-frontend/src/components/DashboardShell.tsx
```diff
@@ -32,30 +32,51 @@ const DashboardShell = ({ user }: DashboardShellProps) => {
...
+        } else {
+          setContributionsCount(
+            contributionsCountResponse?.filter(
+              contribution => contribution?.name !== 'Unassigned',
+            ),
+          );
+        }
+        if (
+          selectedDaos.length === 0 &&
+          !selectedDaos.some(dao => dao.label === 'Unassigned')
```
alexkeating, line 53: Can we make Unassigned a constant? and reuse it if possible

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -535,6 +534,8 @@ export class ContributionCustomResolver {
     const start = args.where.startDate;
     const end = args.where.endDate;
 
+    console.log('args: guildIds', args.where.guildIds)
```
alexkeating, line 537: We probably should remove

alexkeating (commented):
Looks good some small changes

alexkeating:
Also it looks like linting is failing

jonathanprozzi:
> Also it looks like linting is failing

Oh weird I'll take a look. It passed tsc locally

apps/protocol-api/src/prisma/resolvers/Contribution.ts
jonathanprozzi: Done! Missed that one, thanks!

jonathanprozzi:
> Also it looks like linting is failing

Fixed! Forgot I removed a useEffect and left the shell of it still in

alexkeating (approved):

PR #192, opened 2022-09-13
Fix sign out

alexkeating:
## Linear Ticket

## Description

## Screencaptures


apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -727,11 +727,13 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
     }
   };
 
+  console.log(isAuthenticated);
   useEffect(() => {
+    console.log(isAuthenticated);
```
jonathanprozzi, line 732: similar -- can remove

apps/protocol-frontend/src/hooks/useLogout.ts
```diff
@@ -0,0 +1,33 @@
...
+        credentials: 'include',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      });
+      await disconnect();
+    } catch (err) {
+      console.error(err);
+      toast({
+        title: 'Failed to disconnect linear user',
```
jonathanprozzi, line 21: Do we want this to say "Failed to disconnect user" instead of "linear user" ?

apps/protocol-frontend/src/utils/siwe.ts
```diff
@@ -200,7 +200,7 @@ export async function createSiweMessage(
       version: '1',
       chainId: parseInt(chainId),
       nonce: await res.text(),
-      expirationTime: new Date(Date.now() + 1 * 86400000).toISOString(),
+      expirationTime: new Date(Date.now() + 86400000).toISOString(), // 1 day expiration
```
jonathanprozzi, line 203: that's super helpful to know the time value of this -- helpful comment!

apps/protocol-api/src/main.ts
```diff
@@ -395,22 +394,23 @@ app.post('/verify', async function (req, res) {
...
+  } else if (new Date(fields?.data?.expirationTime) <= new Date()) {
     res.status(440).json({ message: 'Token has expired' });
-    return;
   }
 
+  res.end();
+});
+
+app.post('/logout', async function (req, res) {
+  req.session.destroy(() => {
```
jonathanprozzi, line 411: ah! nice -- been testing this out and it's very smooth so far

jonathanprozzi (approved):
This is great! A few small comments for removing console logs but this works really well.
I tested connecting/disconnecting with a single wallet, and also switching between addresses. All work!

One thing for us to consider -- what do you think about redirecting to the `/` upon switch / logging out? I'm not sure what would be expected or preferred UX for this. As it is now it works well if I'm on my Contributions Feed and then want to quickly disconnect/reconnect or switch addresses. In that use case it's nice to be on the same route. Not really sure what's ideal, but wanted to surface for conversation.

Great job!

alexkeating:
> This is great! A few small comments for removing console logs but this works really well. I tested connecting/disconnecting with a single wallet, and also switching between addresses. All work!
> 
> One thing for us to consider -- what do you think about redirecting to the `/` upon switch / logging out? I'm not sure what would be expected or preferred UX for this. As it is now it works well if I'm on my Contributions Feed and then want to quickly disconnect/reconnect or switch addresses. In that use case it's nice to be on the same route. Not really sure what's ideal, but wanted to surface for conversation.
> 
> Great job! firefirefire

Yea, I think not redirecting to start is probably fine for now. There isn't really any benefit from going back to / right?

jonathanprozzi:
> > This is great! A few small comments for removing console logs but this works really well. I tested connecting/disconnecting with a single wallet, and also switching between addresses. All work!
> > One thing for us to consider -- what do you think about redirecting to the `/` upon switch / logging out? I'm not sure what would be expected or preferred UX for this. As it is now it works well if I'm on my Contributions Feed and then want to quickly disconnect/reconnect or switch addresses. In that use case it's nice to be on the same route. Not really sure what's ideal, but wanted to surface for conversation.
> > Great job! firefirefire
> 
> Yea, I think not redirecting to start is probably fine for now. There isn't really any benefit from going back to / right?

Nope, no benefit at all. I think leaving the user on their current page where they disconnect is totally fine

PR #193, opened 2022-09-13
Add Pagination to Contribution

amrro:
## Linear Ticket

[PRO-225](https://linear.app/govrn/issue/PRO-225/paginate-attestation-list-contributions)


apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -390,7 +368,7 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
         });
         await getUserActivityTypes();
         await getUserContributions();
-        await getDaoContributions();
+        // await getDaoContributions();
```
alexkeating, line 371: These can be removed

amrro:
Two issues: 
- For search, we need a separate query that search the db, instead of fetched, paginated items. maybe with different table view!
- Table's headers is not shown up.

apps/contract-sync-job/src/main.ts
```diff
@@ -83,10 +83,12 @@ async function getContribution(
...
-  const contrs = await govrn.contribution.list({
-    where: { on_chain_id: { equals: data.tokenId } },
-    first: 1,
-  });
+  const contrs = (
+    await govrn.contribution.list({
+      where: { on_chain_id: { equals: data.tokenId } },
+      first: 1,
+    })
+  ).result;
```
alexkeating, line 91: It may be good to wrap this in a try catch. I think result will error if list returns an empty

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -52,6 +54,22 @@ type AttestationTableType = {
...
 
+const contributionToTableTypeMapper = (
+  contributions: UIContribution[],
+): AttestationTableType[] =>
+  contributions.map(c => ({
+    id: c.id,
+    date_of_submission: c.date_of_submission,
+    date_of_engagement: c.date_of_submission,
+    attestations: c.attestations,
+    guilds: c.guilds.map((guildObj: any) => guildObj.guild.name)[0] ?? '---',
```
alexkeating, line 65: guiildObj should not be any we should be able to get the type from ui-types since it is on UIContribution

apps/protocol-frontend/src/contexts/ContriubtionContext.tsx
```diff
@@ -0,0 +1,590 @@
...
+import { useNetwork, useSigner } from 'wagmi';
+import { UIAttestations, UIContribution } from '@govrn/ui-types';
+import { networks } from '../utils/networks';
+import { formatDate } from '../utils/date';
+import { ContributionFormValues } from '../types/forms';
+import { GovrnProtocol } from '@govrn/protocol-client';
+import { MintContributionType } from '../types/mint';
+import { UserContext } from './UserContext';
+
+const protocolUrl = import.meta.env.VITE_PROTOCOL_URL;
```
alexkeating, line 15: We should fetch this from a constants file.

apps/protocol-frontend/src/contexts/ContriubtionContext.tsx
```diff
@@ -0,0 +1,590 @@
...
+    >
+      {children}
+    </ContributionContext.Provider>
+  );
+};
+
+type Pagination = {
+  next: () => void;
+  hasMore: boolean;
+};
```
alexkeating, line 541: This might be good to move to a more shared file like in utils

alexkeating (commented):
Looking good, small comments and I am going to click through tomorrow

apps/protocol-frontend/src/contexts/ContriubtionContext.tsx
```diff
@@ -0,0 +1,586 @@
+import { Dispatch, SetStateAction } from 'react';
```
alexkeating, line 1: This filename is mispelled

apps/protocol-frontend/src/contexts/ContriubtionContext.tsx
```diff
@@ -0,0 +1,586 @@
...
+      console.error(error);
+      return null;
+    }
+  };
+
+  const loadNextUserContributionsPage = () => {
+    setUserContributionPage(userContributionPage + 1);
+  };
+
+  const getUserContributions = async (page = 0) => {
```
alexkeating, line 107: I think this needs to wrapped in a useCallback because if userData changes this function will have the original null useData

alexkeating:
Is this still working for you when I pull down I see these for both contributions and attestations

Screenshot from 2022-09-16 12-19-27
Screenshot from 2022-09-16 12-19-35

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -280,20 +286,37 @@ const AttestationsTable = ({
...
-                      {row.cells.map(cell => (
-                        <Td {...cell.getCellProps()} borderColor="gray.100">
-                          {cell.render('Cell')}
-                        </Td>
-                      ))}
-                    </Tr>
-                  );
-                })}
-              </Tbody>
+              <InfiniteScroll
```
alexkeating, line 289: This should wrap the table I am seeing a validate dom nesting error in the console

alexkeating: And same in the attestation table

apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -280,20 +286,37 @@ const AttestationsTable = ({
...
+                dataLength={rows.length}
+                next={() => {
+                  daoContributionPagination.next();
+                }}
+                scrollThreshold={0.8}
```
alexkeating, line 294: Maybe make this threshold .9? Does it feel smoother with that value?

apps/protocol-frontend/src/components/ContributionsTable.tsx
```diff
@@ -329,20 +333,37 @@ const ContributionsTable = ({
...
+            dataLength={rows.length}
+            next={() => {
+              pagination.next();
+            }}
+            scrollThreshold={0.8}
+            hasMore={pagination.hasMore}
+            loader={
+              <tr>
+                <td colSpan={7}>
+                  <GovrnSpinner />
```
alexkeating, line 346: Spinner should be centered in the row on both tables

apps/protocol-frontend/src/components/AttestationsTable.tsx
amrro: It does feel smoother. Also for slower connection, this should be more proactive.

PR #194, opened 2022-09-14
feature/user-daos-query (PRO-402)

jonathanprozzi:
## Linear Ticket

- [PRO-402](https://linear.app/govrn/issue/PRO-402/daos-query-to-show-daos-user-is-in-similar-to-activity-types)

## Description

This adds another query for grabbing the user's DAOs instead of _all_ DAOs. I left the all DAOs query in since we also use this elsewhere in the app.

There is a Set in the DashboardShell where this is used -- in this PR it's only the user's DAOs but when #191 is merged in this Set will include the _Unassigned_ option as well in the array.

## Screencaptures

govrn-dashboard-user-daos


alexkeating (approved):
Looks great, nice and small

PR #195, opened 2022-09-15
Feature/expand proof field (PRO-394)

jonathanprozzi:
## Linear Ticket

- [PRO-394](https://linear.app/govrn/issue/PRO-394/expand-proof-field)

## Description

- This adds the ability for user's to upload a file (under 5mb) to IPFS that will then use the IPFS URI for their Contribution proof.
- Realized we also would need to add this to the Edit form. There are definitely opportunities to refactor these components -- moreso than this feature.

## Screencaptures

https://user-images.githubusercontent.com/9438776/190451122-4c07e148-7026-48c1-b1de-5019e114230b.mp4


apps/protocol-frontend/src/libs/ipfs.ts
```diff
@@ -20,13 +20,23 @@ export const getIPFSClient = () => {
   return ipfs;
 };
 
+export const uploadFileIpfs = async (
+  file: File
+) => {
+  const ipfs = getIPFSClient();
+  const cid = await ipfs.add(file);
+  const resp = await ipfs.pin.add(cid.path);
+  console.log('ipfs resp', resp);
```
alexkeating, line 29: Extra console log

apps/protocol-frontend/src/components/EditContributionForm.tsx
```diff
@@ -89,6 +92,44 @@ const EditContributionForm = ({
...
+        return;
+      }
+    }
+    setSelectedFile(file);
+  };
+
+  const handleIpfsUpload = async () => {
+    if (selectedFile) {
+      setIsUploading(true);
+      const ipfsImageUri = await uploadFileIpfs(selectedFile);
```
alexkeating, line 126: It might be good to wrap this in a try and except and pop a toast if the file fails to be uploaded

apps/protocol-frontend/src/components/ReportForm.tsx
```diff
@@ -50,6 +61,44 @@ const ReportForm = ({ onFinish }: { onFinish: () => void }) => {
...
+
+  const handleIpfsUpload = async () => {
+    if (selectedFile) {
+      setIsUploading(true);
+      const ipfsImageUri = await uploadFileIpfs(selectedFile);
+      setIsUploading(false);
+      setIpfsUri(ipfsImageUri);
+      setValue('proof', ipfsImageUri);
+    }
+  };
```
alexkeating, line 100: Prime candidate  for a hook it looks like. Especially since the logic is duplicated

apps/protocol-frontend/src/components/EditContributionForm.tsx
jonathanprozzi: Oh good call, missed this earlier -- will add this in

apps/protocol-frontend/src/components/ReportForm.tsx
jonathanprozzi: Now that the logic works with the new changes will restructure this

apps/protocol-frontend/src/components/ReportForm.tsx
```diff
@@ -142,14 +208,64 @@ const ReportForm = ({ onFinish }: { onFinish: () => void }) => {
...
+                  size="sm"
+                  leftIcon={<HiCheck />}
+                  variant="outline"
+                  colorScheme="green"
+                  onClick={handleIpfsUpload}
+                  disabled={fileError !== null}
+                >
+                  Upload to IPFS
+                </Button>
+              )} */}
```
alexkeating, line 267: Remove dead code

apps/protocol-frontend/src/components/EditContributionForm.tsx
```diff
@@ -84,9 +80,77 @@ const EditContributionForm = ({
...
+
   const updateContributionHandler: SubmitHandler<
     ContributionFormValues
   > = async values => {
+    if (selectedFile && fileError === null) {
+      try {
+        await uploadFileIpfs(selectedFile, false);
+      } catch (error) {
+        console.error(error);
+        toast({
```
alexkeating, line 144: Should this failure fail adding the contribution?

jonathanprozzi: That's a great question. I wonder if we'd want this to fail since it the user wouldn't be adding in their intended Contribution with a failed proof, but I also can see us allowing it since the proof is optional.

apps/protocol-frontend/src/components/ReportForm.tsx
jonathanprozzi: Oh good catch, thanks!

apps/protocol-frontend/src/components/EditContributionForm.tsx
jonathanprozzi: @alexkeating new behavior: 

https://user-images.githubusercontent.com/9438776/192346457-38afea05-4e85-4ccb-912c-627148055959.mp4

alexkeating (approved):

PR #198, opened 2022-09-19
Rename inputs

alexkeating:
## Description
- Make ids unique


NKoech123 (commented):
This looks really good! 
I wonder if we can resolve the "check annotations" in the unchanged file or have it as technical debt for now? 
Some of the "check annotation" stems from defined variables never being used or that "any" type.

PR #201, opened 2022-09-20
Fix wallet connect issue

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-388/walletconnect-connection-on-rainbowkit-isnt-working

### DIFF
--- a/apps/protocol-frontend/vite.config.ts
+++ b/apps/protocol-frontend/vite.config.ts
@@ -28,6 +28,8 @@ export default defineConfig({
       stream: 'stream-browserify',
       zlib: 'browserify-zlib',
       util: 'util',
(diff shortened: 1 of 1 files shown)


jonathanprozzi:
Having an issue with testing -- when I pull down and run the Rainbow and WC QR codes aren't appearing for me. The Coinbase Wallet one does though. Will keep testing as this is likely on my end

jonathanprozzi (approved):

scottrepreneur (approved):

scottrepreneur:
Wow. Why does it break local??? RIP

alexkeating:
> Woah. Why does it break local??? RIP

Not sure, but I add some logic to use umd in production builds only

PR #205, opened 2022-09-21
 Add Pagination to Contributions Table

amrro:
## Linear Ticket

[PRO-225](https://linear.app/govrn/issue/PRO-225/paginate-attestation-list-contributions) 

## Description
Adding infinite scroll to the contributions table by paginating requests.
This is part of 
- #193


apps/protocol-frontend/src/contexts/ContributionContext.tsx
```diff
@@ -0,0 +1,590 @@
...
+      return null;
+    }
+  };
+
+  const loadNextUserContributionsPage = () => {
+    setUserContributionPage(userContributionPage + 1);
+  };
+
+  const getUserContributions = async (page = 0) => {
+    setUserContributionsLoading(page === 0);
```
alexkeating, line 110: Why is this page === 0 instead of true. Why wouldn't it be loading if we are fetch page 1 or 2 etc...

apps/protocol-frontend/src/contexts/ContributionContext.tsx
```diff
@@ -0,0 +1,590 @@
...
+          title: 'Contribution Report Added',
+          description:
+            'Your Contribution report has been recorded. Add another Contribution report or check out your Contributions.',
+          status: 'success',
+          duration: 3000,
+          isClosable: true,
+          position: 'top-right',
+        });
+        await getUserContributions();
+        await getDaoContributions();
```
alexkeating, line 254: Why remove the activity type fetch?

alexkeating (approved):

apps/protocol-frontend/src/contexts/ContributionContext.tsx
amrro: Good question!
This is for the spinner outside the table. So, if it's set to `true` all time, it will hide the whole table instead. I think this makes no good UX.

PR #208, opened 2022-09-21
Fix/cypress flakiness

NKoech123:
## Linear Ticket

## Description

The goal is to fix flaky tests in Cypress and investigate the code app for potential bugs

**1. Unprepared Element:**  
      DOM element is present but it doesn't behave the way it does in manual testing
   - Root Cause: an element is added to the DOM before the data is ready for it to be interacted with
   - Solution: don't render the element as interactable until it's prepared. e.g disabling
   
**2. Flickering Element:** 
        cy. click() failed because this element is detached from the DOM, but the element is not visible on the page. 
        It's a bug since, on a slower device, the flicker will be more noticeable.
   - Root Cause: the element is added to the DOM, then removed, then a second similar element is added.
                           Declarative rendering
   - Solution:  rework the flow of data so that the element no longer flickers
           : cy.click({force:true}). -but this doesn't fix the flicker

**3. Impatient Test:** 
Test fails without waiting for the condition that would cause it to succeed.
   - Root Cause: The next element to interact with is already present and Cypress doesn't know how to wait
   -  Solution: 
       - Change the app code so the test can proceed immediately
       - Prevent interaction with the element
       - check a "green light"
             - check an element (using` should('be.visible')`)
             - check network request
           
# Actions on failing points
I used the solutions above to fix the flakiness.

Summary:
`command.ts login`- text can be undefined when DOM is detached when fetching `Join Discord`, but can never be undefined when fetching `Connect Wallet`.  Even when undefined when joining Discord, we add `should('be.enabled` condition to wait for the button to be enabled.
                                   
`waitlist test` - form name  had disabled element. Fixed it by increasing timeout and making retries until it's enabled.
`createContribution test` - form name solved same way as explained above.
`updateContribution` - DOM detached solved with the solution expained above.

Resource:  https://www.youtube.com/watch?v=JlBCdRlvPWk

### DIFF
--- a/apps/protocol-frontend-e2e/src/support/commands.ts
+++ b/apps/protocol-frontend-e2e/src/support/commands.ts
@@ -17,13 +17,22 @@ Cypress.Commands.add('login', (address, privateKey) => {
 
   cy.get('button').then($btn => {
     const text = $btn.text();
-    if (text === 'Join Our Discord' || text.slice(0, 2) === '0x') {
-      cy.get('[data-cy="joinOurDiscord-testBtn"]', { timeout: 60000 }).should(
-        'be.visible',
-      );
+    console.log(text);
+    //text can be null/undefined when joining discord
+    if (!text || 
+        typeof text === 'undefined' || 
+        text === 'Join Our Discord' || 
+        text.slice(0, 2) === '0x') {
+      cy.get('[data-cy="joinOurDiscord-testBtn"]', { timeout: 60000 })
+        .should('be.visible');
+
     } else if (text === 'Connect Wallet') {
-      cy.get('[data-cy=connect-wallet]', { timeout: 10000 }).should('be.visible').click();
-      cy.contains('MetaMask').click();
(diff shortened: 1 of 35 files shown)


NKoech123:
Hey Keating, this should be ready for merging now. I fixed the weak points.

apps/protocol-frontend-e2e/src/support/commands.ts
```diff
@@ -17,13 +17,22 @@ Cypress.Commands.add('login', (address, privateKey) => {
...
-    if (text === 'Join Our Discord' || text.slice(0, 2) === '0x') {
-      cy.get('[data-cy="joinOurDiscord-testBtn"]', { timeout: 60000 }).should(
-        'be.visible',
-      );
+    console.log(text);
+    //text can be null/undefined when joining discord
+    if (!text || 
+        typeof text === 'undefined' || 
+        text === 'Join Our Discord' || 
+        text.slice(0, 2) === '0x') {
```
alexkeating, line 25: How can this tect be in so many different states?

PR #210, opened 2022-09-22
feature/split-activity-type-calls (PRO-434)

jonathanprozzi:
## Linear Ticket

- [PRO-434](https://linear.app/govrn/issue/PRO-434/split-out-activity-types-calls)

## Description

- Begins our refactoring to using `react-query` and chip away at our existing `UserContext` by moving our activity types call  to the new hook structure that wraps React Query
- There are a few comments throughout explaining the states and how the lifecycles are used
- Moved our GovrnProtocol initialization into the `UserContext` so we can reuse this throughout
  -  Would appreciate some eyes on this in particular to see if there is a more optimal way - -we may want to consider naming UserContext to something like AppContext? That can be something we consider after the larger refactor
 - Using a Toast in the `main.tsx` that surfaces any errors found in our queries. This will be able to surface the specific error in each query. I made a ticket to pull this into its own component. I'm still tweaking this pattern but will be pretty useful I think

## Screencaptures

Error state example
- This shows the loading state (forcing an error to display the loading state and the error) and an error message in the relevant component and uses a global toast component that surfaces the error in a toast

error-example


apps/protocol-frontend/src/components/ReportForm.tsx
```diff
@@ -109,95 +89,127 @@ const ReportForm = ({ onFinish }: { onFinish: () => void }) => {
...
+  const userActivityTypesHook = useUserActivityTypesList();
+
+  // renaming these on destructuring incase we have parallel queries:
+  const {
+    isLoading: userActivityTypesIsLoading,
+    isFetching: userActivityTypesIsFetching,
+    isError: userActivityTypesIsError,
+    data: userActivityTypesData,
+    error: userActivityTypesError, // unused for now -- handling globally
+  } = userActivityTypesHook;
```
alexkeating, line 101: You can probably combine the destructuring and call but that comes down to preference

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -313,21 +294,21 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
         createWaitlistUser,
         disconnectLinear,
         getAllDaos,
-        isUserActivityTypesLoading,
+        govrnProtocol,
         isUserLoading,
         getUserDaos,
         setAllDaos,
-        setUserActivityTypes,
+        setGovrnProtocol,
```
alexkeating, line 301: I don't think we need this in the values, the context should probably only handle the setting operation

apps/protocol-frontend/src/hooks/useUserActivityTypesList.ts
```diff
@@ -0,0 +1,14 @@
+
+import { useUser } from '../contexts/UserContext';
+import { useQuery } from '@tanstack/react-query';
+import { UIActivityType } from '@govrn/ui-types';
+
+export const useUserActivityTypesList = () => {
+  const { govrnProtocol: govrn } = useUser()
+  const { isLoading, isFetching, isError, error, data } = useQuery(['activityTypes'], async (): Promise<UIActivityType[]> => {
+    const data = await govrn.custom.listActivityTypesByUser({})
```
alexkeating, line 9: Shouldn't the hook take in filter args that resemble listActivityTypesByUser so we don't have to create a new hook for different args?

alexkeating (commented):
Looks really good and simple

apps/protocol-frontend/src/components/ReportForm.tsx
jonathanprozzi: Ah, great idea. I tend to bring them in and then do it, but destructuring upon call will be cleaner especially if/when we have parallel queries. Will be much more readable to do the destructure on call and we'd be able to more directly see the structures.

apps/protocol-frontend/src/hooks/useUserActivityTypesList.ts
jonathanprozzi: Ah, good idea! To make sure I'm on the same page -- we probably want to ensure that the hook's args match the available args on the `listActivityTypesByUser` so that we'd be able to cover all possible use cases in the future? That makes a lot of sense and would be a good pattern for creating these hooks moving forward. This one in particular didn't utilize the args but that's a good pattern to follow for the ones that do. 

Makes sense to support this at a structural level.

jonathanprozzi:
> Looks really good and simple

Awesome! I'm thinking that this more direct and simplified structure scales well and helps us have much more maintainable data flows that's clearer to reason about

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -313,21 +294,21 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
...
         setUserDaos,
         setUserData,
         setUserDataByAddress,
         updateProfile,
-        userActivityTypes,
         userAddress,
         userDaos,
         userData,
         userDataByAddress,
+        useGovrn,
```
alexkeating, line 311: Remove because not necessary

PR #212, opened 2022-09-23
Add Bulk Mint to SDK

amrro:
## Linear Ticket

Part of [PRO-215 - Use bulkMint for bulk flow](https://linear.app/govrn/issue/PRO-215/use-bulkmint-for-bulk-flow)

## Description

Add `bulkMint` to the SDK as a step to support bulk minting in the front-end layer.

## Known Issues
- Transactions Hash is unique in the db and `bulkMint` violates that, since it mints more than one contribution in single transaction. Solved at #213

### DIFF
--- a/libs/protocol-client/src/lib/client/contribution.ts
+++ b/libs/protocol-client/src/lib/client/contribution.ts
@@ -16,8 +16,11 @@ import {
 import { GraphQLClient } from 'graphql-request';
 import { paginate } from '../utils';
 
-const REQUIRED_CONFIRMATIONS = 1;
-const ERROR_CHAIN_ID = 'Failed to fetch on chain Id';
+class ChainIdError extends Error {
+  constructor(message?: string) {
+    super(message || 'Failed to fetch on chain Id');
+  }
+}
 
 export class Contribution extends BaseClient {
   status: ContributionStatus;
@@ -69,7 +72,7 @@ export class Contribution extends BaseClient {
       const transaction = await contract.burnContribution({
         tokenId: contribution.on_chain_id,
       });
-      await transaction.wait(REQUIRED_CONFIRMATIONS);
+      await transaction.wait();
     }
 
     return await this.sdk.deleteContribution({ where: { contributionId: id } });
@@ -111,14 +114,11 @@ export class Contribution extends BaseClient {
       })),
     };
     const transaction = await contract.bulkMint(args);
-    const transactionReceipt = await transaction.wait(REQUIRED_CONFIRMATIONS);
+    const transactionReceipt = await transaction.wait();
 
     const onChainIds: { [index: number]: ethers.BigNumber | null } =
       transactionReceipt.logs
-        .map(l => {
-          console.log('on chain id (logs)', l);
-          return contract.govrn.interface.parseLog(l);
-        })
+        .map(l => contract.govrn.interface.parseLog(l))
         .reduce(
           (prev, current, idx) => ({
             ...prev,
@@ -133,7 +133,7 @@ export class Contribution extends BaseClient {
 
         // TODO: Should we store it anyway as `staged` contribution?
         if (!onChainId) {
-          throw new Error(ERROR_CHAIN_ID);
+          throw new ChainIdError();
         }
 
         if (c.id) {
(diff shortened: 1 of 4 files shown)


libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -16,6 +16,9 @@ import {
 import { GraphQLClient } from 'graphql-request';
 import { paginate } from '../utils';
 
+const REQUIRED_CONFIRMATIONS = 1;
```
alexkeating, line 19: This can be removed, The default argument is 1. We can remove passing in an argument to wait

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -16,6 +16,9 @@ import {
...
+const ERROR_CHAIN_ID = 'Failed to fetch on chain Id';
```
alexkeating, line 20: It may be better to abstract this into a new error type. Something like `class ChainIdError extends Error {
  constructor(message: string) {
    super(message || 'Failed to fetch on chain Id')
}
}`

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -82,6 +85,74 @@ export class Contribution extends BaseClient {
...
+        },
+      })),
+    };
+    const transaction = await contract.bulkMint(args);
+    const transactionReceipt = await transaction.wait(REQUIRED_CONFIRMATIONS);
+
+    const onChainIds: { [index: number]: ethers.BigNumber | null } =
+      transactionReceipt.logs
+        .map(l => {
+          console.log('on chain id (logs)', l);
```
alexkeating, line 119: Remove console log

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -82,6 +85,74 @@ export class Contribution extends BaseClient {
...
+        })
+        .reduce(
+          (prev, current, idx) => ({
+            ...prev,
+            [idx]: current.name === 'Mint' ? current.args['id'] : null,
+          }),
+          {},
+        );
+
+    return await Promise.allSettled(
```
alexkeating, line 130: My worry with this is that we trigger all the requests at once and if there are a lot like 100s then that would put a lot of strain on the backend. It is probably best we we can batch these calls and wait for a batch to finish.

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -180,6 +243,70 @@ export class Contribution extends BaseClient {
...
+    activityTypeId: number;
+    userId: number;
+    args: MintArgs;
+    name: Uint8Array;
+    details: Uint8Array;
+    proof: Uint8Array;
+    txHash: string;
+    onChainId: number;
+  }) {
+    console.log('id in the update:', contribution.id);
```
alexkeating, line 258: Remove console log

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -180,6 +243,70 @@ export class Contribution extends BaseClient {
...
+        ).toString(),
+        proof: ethers.utils.toUtf8String(contribution.proof),
+        status: 'minted',
+        onChainId: contribution.onChainId,
+        userId: contribution.userId,
+        id: contribution.id,
+        txHash: contribution.txHash,
+      },
+    });
+    console.log('update response:', updateResponse);
```
alexkeating, line 277: Remove console log

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -180,6 +243,70 @@ export class Contribution extends BaseClient {
...
+    name: Uint8Array;
+    details: Uint8Array;
+    proof: Uint8Array;
+    txHash: string;
+    onChainId: number;
+  }) {
+    console.log('id in the update:', contribution.id);
+    const updateResponse = this.sdk.updateUserOnChainContribution({
+      data: {
+        name: ethers.utils.toUtf8String(contribution.name),
```
alexkeating, line 261: To make these a little less verbose you can destructure the keys of contribution
`const {args, name, details, proof, onChainId, userId, id, txHash} = contribution`

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -82,6 +85,74 @@ export class Contribution extends BaseClient {
...
+          dateOfSubmission: c.args.dateOfSubmission,
+          dateOfEngagement: c.args.dateOfEngagement,
+        },
+      })),
+    };
+    const transaction = await contract.bulkMint(args);
+    const transactionReceipt = await transaction.wait(REQUIRED_CONFIRMATIONS);
+
+    const onChainIds: { [index: number]: ethers.BigNumber | null } =
+      transactionReceipt.logs
```
alexkeating, line 117: Why not do this in a single for loop? Doesn't map and reduce loop over all the objects in the array twice?

alexkeating (approved):

PR #216, opened 2022-09-27
Pro458 e2e tests/add teardown logic

NKoech123:
## Linear Ticket
https://linear.app/govrn/issue/PRO-458/tests-independence

## Description
- Independent tests....
- add data attributes to react-select components


libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -95,10 +95,12 @@ export class Contribution extends BaseClient {
     proof: Uint8Array,
   ) {
     const contract = new GovrnContract(networkConfig, signer);
+    console.log(contract)
     const transaction = await contract.mint(args);
     const transactionReceipt = await transaction.wait(1);
 
     let onChainId = null as null | ethers.BigNumber;
+    console.log(onChainId)
```
alexkeating, line 103: Extra logs

libs/protocol-ui/src/components/molecules/CreatableSelect/CreatableSelect.tsx
```diff
@@ -6,6 +6,16 @@ import FormLabel from '../../atoms/FormLabel';
 import HelperText from '../../atoms/HelperText';
 import ErrorMessage from '../../atoms/ErrorMessage';
 import { UseFormReturn } from 'react-hook-form/dist/types/form';
+import {components} from 'react-select';
+
+export function SelectContainer(props: any) {
```
alexkeating, line 11: Why is this any? Can we import the react-select types and use it as the type for the prop?

libs/protocol-ui/src/components/molecules/CreatableSelect/CreatableSelect.tsx
```diff
@@ -40,6 +50,7 @@ export interface CreatableSelectProps {
   isDisabled?: boolean;
   variant?: 'outline' | 'filled';
   value?: any;
+  components?: any;
```
alexkeating, line 53: Why is this any? Shouldn't it be like an array of react nodes

libs/protocol-ui/src/components/molecules/Select/Select.tsx
```diff
@@ -45,8 +53,9 @@ export interface SelectProps {
   isDisabled?: boolean;
   variant?: 'outline' | 'filled';
   value?: any;
-
+  innerProps?: any;
```
alexkeating, line 56: Can we be more specific with this type

libs/protocol-ui/src/components/molecules/CreatableSelect/CreatableSelect.tsx
```diff
@@ -6,6 +6,16 @@ import FormLabel from '../../atoms/FormLabel';
 import HelperText from '../../atoms/HelperText';
 import ErrorMessage from '../../atoms/ErrorMessage';
 import { UseFormReturn } from 'react-hook-form/dist/types/form';
+import {components} from 'react-select';
+
+export function SelectContainer(props: any) {
+  return (
+     <components.SelectContainer
+        {...props}
+        innerProps={Object.assign({}, props.innerProps, { 'data-cy': 'daoCreatableSelect-testing' })}
```
jonathanprozzi, line 15: We'll want to make this passed in via props optionally instead of hardcoded -- this'll result in every CreatableSelect having that data-id

libs/protocol-ui/src/components/molecules/Select/Select.tsx
```diff
@@ -8,6 +8,14 @@ import FormLabel from '../../atoms/FormLabel';
 import HelperText from '../../atoms/HelperText';
 import ErrorMessage from '../../atoms/ErrorMessage';
 
+export function SelectContainer(props: any) {
+  return (
+     <components.SelectContainer
+        {...props}
+        innerProps={Object.assign({}, props.innerProps, { 'data-cy': 'daoSelect-testing' })}
```
jonathanprozzi, line 15: We'll want to make this passed in via props optionally instead of hardcoded -- this'll result in every Select having that data-id

apps/protocol-frontend-e2e/src/integration/04_updateContribution.cy.ts
```diff
@@ -17,14 +20,19 @@ beforeEach(() => {
   cy.get('[data-cy="contributionsSidebar-btn"]', { timeout: 15000 })
     .should('be.visible')
     .click({ force: true });
+ 
+  cy.get(".infinite-scroll-component")
```
jonathanprozzi, line 24: Nice!

libs/protocol-ui/src/components/molecules/CreatableSelect/CreatableSelect.tsx
jonathanprozzi, line 11: We should be able to pass this in as a prop from the parent CreatableSelect

jonathanprozzi:
@NKoech123 after our sync earlier I tried out a few things. This change resolves the type issue so that the props are typed and then also allows for this to be done when needed. It's still not exactly what we talked about -- it would be preferable to pass in `dataId` as a prop. We can keep playing with that, but check this out:

```jsx

import React from 'react';
import { InputProps } from 'react-select';
import Creatable from 'react-select/creatable';
import { Controller } from 'react-hook-form';
import { Box, FormControl, Stack } from '@chakra-ui/react';
import FormLabel from '../../atoms/FormLabel';
import HelperText from '../../atoms/HelperText';
import ErrorMessage from '../../atoms/ErrorMessage';
import { UseFormReturn } from 'react-hook-form/dist/types/form';
import { components } from 'react-select';

type Errors = {
  [name: string]: {
    label?: {
      message?: string;
    };
    value?: {
      message?: string;
    };
    message: string;
  };
};

type Option = {
  label: string;
  value: string;
};

const CustomInput = (props: InputProps) => {
  return (
    <components.Input {...props} data-testid="daoCreatableSelect-testing" />
  );
};

export interface CreatableSelectProps {
  name: string;
  label?: string;
  placeholder?: string;
  defaultValue?: Option | Option[];
  id?: string;
  tip?: string;
  options: Option[];
  isRequired?: boolean;
  localForm: Pick<UseFormReturn, 'control' | 'formState'>;
  isMulti?: boolean;
  isClearable?: boolean;
  onChange?: (option: Option) => void;
  isDisabled?: boolean;
  variant?: 'outline' | 'filled';
  useCustomInput?: boolean;
}

const CreatableSelect: React.FC<CreatableSelectProps> = ({
  label,
  name,
  tip,
  placeholder,
  defaultValue,
  options,
  isMulti,
  isClearable,
  onChange,
  isDisabled,
  localForm,
  useCustomInput = false,
}: CreatableSelectProps) => {
  const {
    control,
    formState: { errors },
  } = localForm;

  return (
    <FormControl mb={4} isInvalid={errors && errors[name] !== undefined}>
      <Stack spacing={2}>
        {label && <FormLabel label={label} />}
        {tip && <HelperText tipText={tip} fontSize="xs" color="gray.700" />}
        <Box my={2}>
          <Controller
            name={name}
            control={control}
            shouldUnregister={false}
            render={({ field }) => (
              <Creatable
                {...field}
                onBlur={field.onBlur}
                options={options}
                defaultValue={defaultValue}
                placeholder={placeholder}
                isClearable={isClearable}
                isMulti={isMulti}
                onChange={onChange}
                isDisabled={isDisabled}
                components={useCustomInput ? { Input: CustomInput } : {}}
              />
            )}
          />
          {errors && (
            <ErrorMessage
              errorMessage={errors[name] && errors[name]?.message?.toString()}
            />
          )}
        </Box>
      </Stack>
    </FormControl>
  );
};

export default CreatableSelect;

```

apps/protocol-frontend-e2e/src/support/commands.ts
```diff
@@ -41,4 +41,140 @@ Cypress.Commands.add('interceptGQL', (httpMethod, operationNames) => {
...
+      cy.task('queryDatabase', GuildUser1Query );
+  }
+
+});
+
+// tearDown DB
+Cypress.Commands.add('teardownDB',(tableNames)=>{
+    for (const tableName of tableNames){
+      const teardownQuery = `
+        TRUNCATE TABLE "${tableName}" RESTART IDENTITY CASCADE;
```
alexkeating, line 166: We don't want to truncate the database. This will delete data that was created outside of the tests. Instead we want to be very specific with what we delete. And it is okay to have a different teardown for each test. We don't need a general teardown function

apps/protocol-frontend-e2e/src/support/commands.ts
```diff
@@ -41,4 +41,140 @@ Cypress.Commands.add('interceptGQL', (httpMethod, operationNames) => {
       }
     });
   }
+
+});
+
+// Seed DAOs/Guilds into DB
+Cypress.Commands.add('seedDB',(tableName)=>{
```
alexkeating, line 48: This command is probably doing too much it would probably be better to structure this. Instead of using a command it is probably better to create functions for such as create_guild() or create_contribution that are called in the before each hook. And then have a separate delete_guild or delete_contribution function.

apps/protocol-frontend-e2e/src/support/commands.ts
NKoech123: Yeah, this makes sense! Will fix this in a moment

apps/protocol-frontend-e2e/src/support/commands.ts
NKoech123: I see.  Just to clarify, I create a `**.ts` file under path `protocol-frontend-e2e/src/support` where I can have these functions?

libs/protocol-ui/src/components/molecules/Select/Select.tsx
NKoech123: @alexkeating @jonathanprozzi this might have to be a debt.

alexkeating (approved):

PR #229, opened 2022-09-30
Downgrade siwe

alexkeating:
## Linear Ticket

### DIFF
--- a/apps/protocol-api/src/main.ts
+++ b/apps/protocol-api/src/main.ts
@@ -4,7 +4,7 @@ import express from 'express';
 import Session from 'cookie-session';
 import { PrismaClient } from '@prisma/client';
 import { applyMiddleware } from 'graphql-middleware';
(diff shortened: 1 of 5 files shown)


jonathanprozzi (approved):
Thinking that the recent upgrade to 2 is causing issues on staging?

jonathanprozzi:
I think we can switch out `SiweErrorType` with `ErrorTypes` -- this resolves the build issue for the `SiweErrorType`
Looking for the message issue now

alexkeating:
> Thinking that the recent upgrade to 2 is causing issues on staging?

Could be I am testing it out

jonathanprozzi:
Cool -- I was about to suggest validate instead of verify. Came across that too

Approved already. We can merge once CI is solid

alexkeating:
> Cool -- I was about to suggest validate instead of verify. Came across that too
> 
> Approved already. We can merge once CI is solid

Yea, thanks I saw the errors in the build

PR #242, opened 2022-10-11
Split out bulk mint contribution

amrro:
## Linear Ticket

[PRO-486 - Split out bulk mint contribution](https://linear.app/govrn/issue/PRO-486/split-out-bulk-mint-contribution)

## Description
Part of migrating to react-query from using Context.

### DIFF
--- a/apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
+++ b/apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
@@ -32,10 +32,10 @@ const useContributionBulkMint = () => {
   const { userData, govrnProtocol: govrn } = useUser();
 
   const { mutateAsync, isLoading, isError, isSuccess } = useMutation(
-    ['bulk-mint-contribution'],
+    ['BulkMintContribution'],
     async (contributions: BulkMintOptions[]) => {
       if (!(signer && chain?.id && userData)) {
-        throw new Error('What should the message be!?');
+        throw new Error('Not signed in, unable to mint');
       }
 
       return await govrn.contribution.bulkMint(
@@ -62,10 +62,10 @@ const useContributionBulkMint = () => {
       );
     },
     {
-      onSuccess: result => {
+      onSuccess: async result => {
         /* Failed writes/updates to db are included in result. */
 
-        const minted = result.filter(i => i.status === 'fulfilled');
+        const minted = result.filter(promise => promise.status === 'fulfilled');
 
         if (minted.length > 0) {
           queryClient.invalidateQueries(['contributionList']);
@@ -82,8 +82,11 @@ const useContributionBulkMint = () => {
         }
 
         const failedToMint = result
-          .filter((i): i is PromiseRejectedResult => i.status === 'rejected')
-          .filter(i => i.reason instanceof ChainIdError);
+          .filter(
+            (promise): promise is PromiseRejectedResult =>
+              promise.status === 'rejected',
+          )
+          .filter(rejected => rejected.reason instanceof ChainIdError);
 
         if (failedToMint.length > 0) {
           toast.error({
--- a/apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
+++ b/apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
@@ -1,0 +1,114 @@
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+import useGovrnToast from '../components/toast';
+import { useNetwork, useSigner } from 'wagmi';
+import { useUser } from '../contexts/UserContext';
+import { networks } from '../utils/networks';
+import { ethers } from 'ethers';
+import { ChainIdError } from '@govrn/protocol-client';
+import pluralize from 'pluralize';
+
+type BulkMintOptions = {
+  ipfsContentUri: string;
+  engagementDate: Date | string;
+  attestations: { id: number }[];
+  guildName: string;
+  name: string;
+  action: string;
+  details?: string | null;
+  id: number;
+  proof?: string | null;
+  txHash?: string | null;
+  user: { id: number };
+  date_of_submission: Date | string;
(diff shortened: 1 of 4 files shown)


apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const useContributionBulkMint = () => {
+  const toast = useGovrnToast();
+  const queryClient = useQueryClient();
+  const { data: signer } = useSigner();
+  const { chain } = useNetwork();
+  const { userData, govrnProtocol: govrn } = useUser();
+
+  const { mutateAsync, isLoading, isError, isSuccess } = useMutation(
+    ['bulk-mint-contribution'],
```
alexkeating, line 35: We are using title case for keys

apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
```diff
@@ -0,0 +1,114 @@
...
+    async (contributions: BulkMintOptions[]) => {
+      if (!(signer && chain?.id && userData)) {
+        throw new Error('What should the message be!?');
```
alexkeating, line 38: See the mint pr for copy to work off of

apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
```diff
@@ -0,0 +1,114 @@
...
+    },
+    {
+      onSuccess: result => {
+        /* Failed writes/updates to db are included in result. */
+
+        const minted = result.filter(i => i.status === 'fulfilled');
+
+        if (minted.length > 0) {
+          queryClient.invalidateQueries(['contributionList']);
+          queryClient.invalidateQueries(['contributionInfiniteList']);
```
alexkeating, line 72: Should we be awaiting these

apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
```diff
@@ -0,0 +1,114 @@
...
+              'Contribution',
+              minted.length,
+              true,
+            )} has been minted.`,
+          });
+        }
+
+        const failedToMint = result
+          .filter((i): i is PromiseRejectedResult => i.status === 'rejected')
+          .filter(i => i.reason instanceof ChainIdError);
```
alexkeating, line 86: Probably good to be more descriptive than i for the variable names

apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
```diff
@@ -0,0 +1,114 @@
...
+            description: `${pluralize(
+              'Contribution',
+              minted.length,
+              true,
+            )} has been minted.`,
+          });
+        }
+
+        const failedToMint = result
+          .filter((i): i is PromiseRejectedResult => i.status === 'rejected')
```
alexkeating, line 85: Why do we need to explicit about the type here and not line 68

amrro: `filter` returns the same type as its input.
For successful mints, we only care about the count. 
for failed mints, we are interested in the `PromiseRejectedResult.reason`. Consequently, we need to narrow down the type that's why I added the type predicate `i is PromiseRejectedResult`


I found out that there is narrowing-down [`filter`](https://github.com/microsoft/TypeScript/blob/4f54e7e947298162d29f3104265e74dcfbc90d82/lib/lib.es5.d.ts#L1238-L1243) But it works with inheritance, Not sure if this applies to type union also or not.

amrro: Both ways are not perfectly safe!
```ts
const failedToMint = result
  .filter<PromiseRejectedResult>((i): i is PromiseRejectedResult => {
                                   // ^ forces to put type predicate! (bonus)
    return i.status === 'rejected';
        // ^ You can change the predicate, but it won't catch that it's different type.
})
```

apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
amrro: - I think this is kinda of fire-and-forget promise. We don't care about the result as much as triggering fetching it. 
- It would propagate errors of these queries here, I made example [here](https://www.typescriptlang.org/play?#code/MYewdgzgLgBADgJxAWwJYQKYAYCMMC8MAhhAJ5jAwAUAlAQHwwDeMAUDDKJCADYYB0PEAHMqAIgBUAOQDKATSkBhGAFoYAFQAW6GAEkYABSRpMMOSACuYmgG52MBBigWEYQ8fQDHEXgDcM4tLySqoa2hB67iieZpbWdgC+rKxc0PAemFgATATEZBTUdPiMLPapvAIYCEgI4lIgUSYYEQBGpDAymkTkGDAA6kQIACbx9lCaSADuMGAY0wCi1SC1YvWNnq3tnd2z-YMjtqxJKeBpAGYWPDxnqFcAsr2EJOSUtAzMbBxEk0SosIjRTI4Wh2L4-P7pQHYLIg+yOZyuGBiPpdKAYfwIACEYiOyQuVxu9wCNH44wwYCovne5T4ghElJoNFY-GARCgwE0VEejBpAiEojEAHkAKrqGS6AAi8wAXGIADQwDCM5KsIA)

alexkeating: Cool, yea agreed

apps/protocol-frontend/src/hooks/useContributionBulkMint.ts
alexkeating: Ah I, think the easiest way to narrow down the type is to use a for loop and rely on type inference. I am always wary with type hinting because it tends to cause problems eventually

alexkeating (approved):

PR #244, opened 2022-10-12
Feature/add chain to contribution

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-516/contribution-on-chain-id-should-be-unique-by-chain

## Description

* Creating contributions on multiple chains is causing issues goerli vs Gnosis. We should tie every minted contribution to a chain


apps/protocol-api/src/prisma/resolvers/Attestation.ts
```diff
@@ -135,7 +138,10 @@ export class AttestationResolver {
         },
         contribution: {
           connect: {
-            on_chain_id: args.data.contributionOnChainId,
+            chain_id_on_chain_id: {
```
Flip-Liquid, line 141: @alexkeating it doesn't look like this lines up with the fields for `ContributionWhereUniqueInput`

PR #248, opened 2022-10-13
Add `MintAttestation` Hook

amrro:
## Linear Ticket
[PRO-520 - Remove mintAttestaion](https://linear.app/govrn/issue/PRO-520/remove-mintattestaion)

## Description
Add `MintAttestation` Hook as part of migrating to use `react-query`

### DIFF
--- a/apps/protocol-frontend/src/types/table.ts
+++ b/apps/protocol-frontend/src/types/table.ts
@@ -1,7 +1,5 @@
 import { ReactNode } from 'react';
 
-type Status = 'minted' | 'staging' | string;
-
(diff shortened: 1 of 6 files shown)


apps/protocol-frontend/src/types/table.ts
```diff
@@ -1,5 +1,7 @@
 import { ReactNode } from 'react';
 
+type Status = 'minted' | 'staging' | string;
```
alexkeating, line 3: This doesn't narrow down the type. We might as well keep it as a string

alexkeating (approved):

PR #252, opened 2022-10-14
fix/bulk-attribute-updating (PRO-518)

jonathanprozzi:
## Linear Ticket

- [PRO-518](https://linear.app/govrn/issue/PRO-518/bulk-attribute-to-dao-not-updating)

## Description

- Fixes issue with the `BulkDaoAttribute` where there were missing values required in the update mutation 
- Passes in the `id` into the GovrnToast so we only show a single Toast on multiple successes 

## Screencaptures

bulk-attribute-dao-update


apps/protocol-frontend/src/components/ContributionsTable.tsx
```diff
@@ -104,11 +103,15 @@ const ContributionsTable = ({
           id: contribution.id,
           details: contribution.details,
           proof: contribution.proof,
+          updatedAt: contribution.updatedAt,
           date_of_submission: contribution.date_of_submission,
           engagementDate: formatDate(contribution.date_of_engagement),
+          date_of_engagement: formatDate(contribution.date_of_engagement),
```
jonathanprozzi, line 109: I know we have some redundancy here. We should aim to refactor these to match -- this is from before we were strict with typing.

apps/protocol-frontend/src/components/BulkDaoAttributeModal.tsx
```diff
@@ -56,13 +55,16 @@ const BulkDaoAttributeModal = ({
     },
   ];
 
+  console.log('contributions', contributions);
```
alexkeating, line 58: Leftover log

apps/protocol-frontend/src/components/BulkDaoAttributeModal.tsx
```diff
@@ -56,13 +55,16 @@ const BulkDaoAttributeModal = ({
...
   const combinedDaoListOptions = [...new Set([...daoReset, ...daoListOptions])];
 
   const bulkAttributeDaoHandler: SubmitHandler<
     BulkDaoAttributeFormValues
-  > = async (values: BulkDaoAttributeFormValues) => {
+  > = async values => {
     setAttributing(true);
-    contributions.map((contribution, idx) => {
+    contributions.map(contribution => {
+      console.log('contribution in map', contribution);
```
alexkeating, line 67: Left over log

alexkeating (approved):

PR #255, opened 2022-10-18
feature/dao-dashboard-auth-wrapper (PRO-526)

jonathanprozzi:
## Linear Ticket

- [PRO-526]()

## Description

Putting this up for a quick review. I included two approaches that both work but wanted some feedback on which we prefer. 

This adds a `userDaos` to the UserContext without an extra call since we have that in the `userData` and then there are two approaches:

1) a `useUserDaoMember` hook that has a guildId passed in and returns whether or not a user is part of the DAO

2) a `isUserDaoMember` call in context that does the same

I like the hook approach more, but both work and so I wanted to put up for feedback. I'll remove the option we don't use so the commented code will be removed before merging

This then checks based on the `guildId` in the route param and is structured similar to the  `<RequireActiveUser>` wrapper in Routes

## Screencaptures

Authorized (will be the DAO Dashboard UI):

image

Unauthorized (can also redirect instead):

image


apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -87,6 +95,19 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
...
   }, [address]);
 
+  const getUserDaos = useCallback(() => {
+    if (userData) {
+      const userDaos = userData?.guild_users;
+      setUserDaos(userDaos);
+    }
+  }, [userData]);
+
+  const isUserDaoMember = (daoId: string): boolean | undefined => {
```
jonathanprozzi, line 105: This'll be removed if we prefer the hook

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -199,19 +220,27 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
     }
   }, [userDataByAddress, isAuthenticated]);
 
+  useEffect(() => {
```
jonathanprozzi, line 223: This useEffect call is necessary since it relies on `userData` to exist -- tried to do without a useEffect since that would be more optimal

apps/protocol-frontend/src/hooks/useUserDaoMember.ts
```diff
@@ -0,0 +1,9 @@
+import { useUser } from "../contexts/UserContext";
+
+
+export const useUserDaoMember = (daoId: string): boolean | undefined => {
+  const { userDaos } = useUser();
+  if (userDaos !== null && daoId !== undefined) {
+    return userDaos.map(dao => dao.guild_id).includes(parseInt(daoId));
```
amrro, line 7: This doesn't utilize `useQuery`.

amrro (commented):
IMHO, if we don't use react queries, it would be more contextual to include it within `UserContext`. 

What I am trying to say is this is not really a hook, it's just a simple function.

apps/protocol-frontend/src/hooks/useUserDaoMember.ts
jonathanprozzi: Yup, makes sense. I think the reasoning to include in UserContext is solid -- I was on the fence but that was part of my consideration. Since it's not really a hook and it does seem to make more sense in Context.

jonathanprozzi:
> IMHO, if we don't use react queries, it would be more contextual to include it within `UserContext`.
> 
> What I am trying to say is this is not really a hook, it's just a simple function.

I think that's a great point overall. This definitely does seem to make more sense in UserContext for all those reasons.

I can remove the hook and return to the one in UserContext

Thanks for the feedback!

apps/protocol-frontend/src/contexts/UserContext.tsx
jonathanprozzi: Ah! Sounds good. I originally had it as a value from mapping over the values coming from user data but didn't use a Map. I'll make that change and keep all of this in context and remove the extra hook. Thanks for the feedback!

apps/protocol-frontend/src/pages/DaoDashboard.tsx
```diff
@@ -0,0 +1,63 @@
...
+      >
+        <Box
+          background="white"
+          boxShadow="sm"
+          borderRadius={{ base: 'none', md: 'lg' }}
+        >
+          {isConnected && isAuthenticated && userData ? (
+            <UserView />
+          ) : (
+            <NewUserView />
```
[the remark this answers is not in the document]
jonathanprozzi: Yup - the next PR for the name header removes the placeholders

alexkeating (approved):

PR #256, opened 2022-10-20
Review Contract Sync Job

amrro:
## Linear Ticket
[PRO-535 - Verify contract sync job is working](https://linear.app/govrn/issue/PRO-535/verify-contract-sync-job-is-working)

## Description
Review that Contract sync Job is working: 

### Contributions: 
- [x] Load all contributions Events from Subgraph. 
- [x] Insert loaded contribution into the database:
    - [ ] Successfully load from Details URI.
    - [x] Map data out to be inserted into db. 
    - [x] Insert them into the database

### Attestations: 
- [x] Load all attestations Events from Subgraph. 
- [x] Insert loaded contribution into the database:


apps/contract-sync-job/src/db-helpers.ts
```diff
@@ -0,0 +1,73 @@
+import { GovrnProtocol } from '@govrn/protocol-client';
+
+export const getOrInsertUser = async (
+  govrn: GovrnProtocol,
+  data: { address: string },
+): Promise<number> => {
+  // Query existing user.
+  const user = await govrn.user.list({
```
alexkeating, line 8: Could we use get instead of list?

apps/contract-sync-job/src/db-helpers.ts
```diff
@@ -0,0 +1,73 @@
+import { GovrnProtocol } from '@govrn/protocol-client';
```
alexkeating, line 1: Maybe omit the -helpers for the file name and call db?

apps/contract-sync-job/src/db-helpers.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  return newUser.id;
+};
+
+export const getOrInsertActivityType = async (
+  govrn: GovrnProtocol,
+  data: { name: string },
+): Promise<number> => {
+  // Check for an existing activity types.
+  const activityTypeId = await govrn.activity_type.list({
```
alexkeating, line 34: Same here is get possible over list?

apps/contract-sync-job/src/main.ts
```diff
@@ -8,123 +7,43 @@ import {
...
+import { fetchIPFS } from './utils';
+import {
+  createJobRun,
+  getContribution,
+  getOrInsertActivityType,
+  getOrInsertUser,
+} from './db-helpers';
 
-const protcolUrl = process.env.PROTOCOL_URL;
+const PROTOCOL_URL = process.env.PROTOCOL_URL;
```
alexkeating, line 18: Nice

apps/contract-sync-job/src/main.ts
```diff
@@ -8,123 +7,43 @@ import {
...
-const jobName = 'contract-sync-job';
+const JOB_NAME = 'contract-sync-job';
 
 const CONTRACT_ADDRESS = process.env.CONTRACT_ADDRESS;
 const CHAIN_URL = process.env.CHAIN_URL;
+
 const networkConfig: NetworkConfig = {
   address: CONTRACT_ADDRESS,
-  chainId: 2,
+  chainId: 5,
```
alexkeating, line 28: Why are we chaning the chainId here?

apps/contract-sync-job/src/utils.ts
```diff
@@ -0,0 +1,31 @@
+import fetch from 'node-fetch';
```
alexkeating, line 1: Might be best to rename this file to ipfs? and  remove the logger since it isn't being used

apps/contract-sync-job/src/main.ts
```diff
@@ -149,9 +68,7 @@ const main = async () => {
         });
 
         const detailsUri = ethers.utils.toUtf8String(contr.detailsUri);
-        const ipfsBlob = await fetchIPFS(detailsUri);
-        const ipfsText = await ipfsBlob.text();
-        const contributionDetails = JSON.parse(ipfsText);
+        const contributionDetails = await fetchIPFS(detailsUri);
```
alexkeating, line 71: Can't this error out and stop the job? We want to make sure this error can be handled gracefully and the job can continue processing. It might be good to catch the parse error here continue processing

apps/contract-sync-job/src/db-helpers.ts
amrro: `govrn.user.get` uses only `id` number to query.
In this case, we want to query by unique address.

apps/contract-sync-job/src/db-helpers.ts
amrro: `ActivityTypes` Resolvers doesn't implement a `get()` function. 
Similar to `user` we query by `name`.

apps/govrn-contract-subgraph/build/subgraph.yaml
```diff
@@ -4,10 +4,10 @@ schema:
 dataSources:
   - kind: ethereum
     name: Govrn
-    network: rinkeby
+    network: xdai
```
alexkeating, line 7: These changes are residual from redeploying the gnosis subgraph

apps/contract-sync-job/src/main.ts
```diff
@@ -149,32 +68,37 @@ const main = async () => {
...
+            activity_type_id: contributionActivityTypeId,
+            user_id: userId,
+            date_of_engagement: new Date(contr.dateOfEngagement.toNumber()),
+            date_of_submission: new Date(contr.dateOfSubmission.toNumber()),
+            details: contributionDetails.details,
+            proof: contributionDetails.proof,
+            on_chain_id: Number(event.id),
+          };
+        } catch {
+          return null;
```
alexkeating, line 85: We return null here and filter rather than an object with empty values

apps/contract-sync-job/src/ipfs.ts
```diff
@@ -0,0 +1,35 @@
...
+}> => {
+  const auth =
+    'Basic ' +
+    Buffer.from(
+      process.env.INFURA_PROJECT_ID + ':' + process.env.INFURA_PROJECT_SECRET,
+    ).toString('base64');
+
+  const resp = await fetch(getIpfsFromHash(ipfsHash), {
+    headers: {
+      Authorization: auth,
```
alexkeating, line 24: Also, I noticed things weren't working as expected with the custom domain. We needed auth and a different uri path

PR #258, opened 2022-10-20
get guild contribution count by activity type

Flip-Liquid:
## Linear Ticket

[Pro-530](https://linear.app/govrn/issue/PRO-530/create-an-endpoint-for-getting-the-contributions-by-type)

## Description

Create a new query for retrieving a count of contributions to a particular dao in a particular date range frouped by activity id and activity name.

### DIFF
--- a/apps/protocol-api/src/main.ts
+++ b/apps/protocol-api/src/main.ts
@@ -71,6 +71,7 @@ const permissions = shield(
       guilds: or(isAuthenticated, hasToken),
       listUserByAddress: isAuthenticated,
       getContributionCountByDateForUserInRange: or(isAuthenticated, hasToken),
+      getContributionCountForDaoByActivityTypeInRange: or(isAuthenticated, hasToken),
       users: hasToken,
       jobRuns: hasToken,
       linearUsers: hasToken,
@@ -78,6 +79,7 @@ const permissions = shield(
       contributionStatuses: hasToken,
     },
     ContributionCountByDate: or(isAuthenticated, hasToken),
+    ContributionCountByActivityType: or(isAuthenticated, hasToken),
     Mutation: {
       '*': deny,
       createActivityType: hasToken,
@@ -475,4 +477,5 @@ app.get('/linear/oauth', async function (req, res) {
   }
 });
 
+console.log('listening');
 app.listen(4000);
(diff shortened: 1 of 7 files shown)


apps/protocol-api/src/main.ts
```diff
@@ -71,13 +71,15 @@ const permissions = shield(
       guilds: or(isAuthenticated, hasToken),
       listUserByAddress: isAuthenticated,
       getContributionCountByDateForUserInRange: or(isAuthenticated, hasToken),
+      getContributionCountForDaoByActivityTypeInRange: or(isAuthenticated, hasToken),
```
alexkeating, line 74: Would it be better to make this a more generic name. Like getActivityTypeContributionCount. And have the dao part of the name and in range part be arguments set by the client so we want reuse the logic. Trying to avoid having a bunch of different methods for slightly different functionality

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -226,6 +226,26 @@ export class GetUserContributionCountArgs {
   where!: GetContributionCountForUser;
 }
 
+@TypeGraphQL.InputType('GetDaoContributionCountInput')
+export class GetDaoContributionCount {
+  @TypeGraphQL.Field(_type => Number)
+  daoId: number;
```
alexkeating, line 232: Maybe change this as guildId for consistency with other naming in the db

apps/protocol-api/src/main.ts
Flip-Liquid: good idea

Flip-Liquid: Just renaming for the purposes of this PR. added Pro-551 for making the logic totally generic (tolerating nullable fields)

libs/protocol-client/src/lib/graphql/queries.graphql
```diff
@@ -445,10 +445,10 @@ query getContributionCountByDateForUserInRange(
   }
 }
 
-query getContributionCountForDaoByActivityTypeInRange(
+query getContributionCountByActivityType(
   $where: GetDaoContributionCountInput!
```
jonathanprozzi, line 449: Should this be `GetContributionInput` instead of `GetDaoContributionCountInput` due to the name change? I was getting an error and noticed the different name

jonathanprozzi (commented):
Looks good other than the question I had about the name for the Input!

Flip-Liquid:
@jonathanprozzi The changes to make the input generic are in another PR (#262). mind if we take this feedback in that one?

jonathanprozzi:
> @jonathanprozzi The changes to make the input generic are in another PR (#262). mind if we take this feedback in that one?

Ah got it! I was testing it on this one and ran into an error with generating the types, but that makes sense! 

I think this looks good then

PR #259, opened 2022-10-21
feature/dao-dashboard-dao-header (PRO-476)

jonathanprozzi:
## Linear Ticket

- [PRO-476](https://linear.app/govrn/issue/PRO-476/dao-name-header)

## Description

- Adds in the Header and general page layout for the DAO Dashboard
- Added the `guild` object with `id` and `name` to what's returned in the `guild_users` in `userDaos` so we have access to those -- will also see us wanting to use this at a later point to display a user's DAOs 
- Other PRs will continue to remove the placeholders on this page


apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -1,29 +1,105 @@
...
+        <Flex direction="column" gap={4}>
+          <Flex direction={{ base: 'column' }} gap={2}>
+            <Flex
+              direction={{ base: 'column', lg: 'row' }}
+              gap={8}
+              justifyContent="space-apart"
+            >
+              {/* <Flex flexBasis="50%"> */}
+              {/* DatePicker will go here */}
+              {/* </Flex> */}
```
alexkeating, line 36: Extra commented code

alexkeating (approved):

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
jonathanprozzi: Will remove in next PR - no need to have this in since it's only indicating where the component will go. Thanks!

PR #260, opened 2022-10-21
feature/dao-dashboard-recent-contributions-table (PRO-479)

jonathanprozzi:
## Linear Ticket

- [PRO-479](https://linear.app/govrn/issue/PRO-479/[table]-recent-contribution-list)

## Description

- Added the Recent Contributions Table and supporting type RecentContributionsTableType
- This is for the table itself -- added it to the DaoDashboardShell UI but we'll want to tweak the whole layout once all of the components are in
- There is a commented out placeholder for the Flex wrapper on the Shell -- leaving this in since we'll probably need it once these are all done
  - I made a new ticket for the layout adjustments after these are all in

## Screencaptures

dao-dashboard-recent-contributions


apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -23,81 +25,45 @@ const DaoDashboardShell = ({ daoName }: DaoDashboardShellProps) => {
...
+      > */}
+      <Flex direction="column" gap={4}>
+        <Flex direction={{ base: 'column' }} gap={2}>
+          <Flex
+            direction={{ base: 'column', lg: 'row' }}
+            gap={8}
+            justifyContent="space-apart"
+          ></Flex>
+          <Flex direction="column" gap={2}>
+            <Heading as="h3" fontSize="lg" color="gray.800" fontWeight="normal">
```
alexkeating, line 37: Could fontSize, color and fontWeight be passed on the top level div so the props are repeated

alexkeating (approved):
Looks good, suggested a few small improvements

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
jonathanprozzi: Great idea! We should do that more often generally. We do it at some parts

jonathanprozzi:
> Looks good, suggested a few small improvements

Thanks!

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
jonathanprozzi: I made this change, but it had some side effects -- as we tweak the actual layout for this page I'll optimize this

PR #262, opened 2022-10-24
add getContributionCount endpoint

Flip-Liquid:
## Linear Ticket
[Pro-527](https://linear.app/govrn/issue/PRO-527/create-a-getdaocontributioncount-endpoint)

## Description

Add `getContributionCount` endpoint 

## Screencaptures

### DIFF
--- a/apps/contract-sync-job/src/db.ts
+++ b/apps/contract-sync-job/src/db.ts
@@ -1,0 +1,73 @@
+import { GovrnProtocol } from '@govrn/protocol-client';
+
+export const getOrInsertUser = async (
+  govrn: GovrnProtocol,
+  data: { address: string },
+): Promise<number> => {
+  // Query existing user.
+  const user = await govrn.user.list({
+    where: {
+      address: { equals: data.address },
+    },
+    first: 1,
+  });
+
+  // Returns the id, if user exists.
+  if (user.length === 1) {
+    return user[0]?.id;
+  }
+
+  // Insert new user into db, if user doesn't exist.
+  const newUser = await govrn.user.create({
+    address: data.address,
+    username: data.address,
+  });
+
+  return newUser.id;
+};
+
+export const getOrInsertActivityType = async (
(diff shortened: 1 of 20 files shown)


jonathanprozzi:
@Flip-Liquid nice! Works well for getting the total Contribution count by DAO. Passing in the filters as you described does exactly this.

Question I have while also thinking about PRO-478 (Contribution percent by Member in a Date Range)

If we change the return on this from an Int to either the ContributionFragment or something that includes the user (with id and name) we'd be able to use this for the pie chart showing Member % / Total %.

What do you think?

Flip-Liquid:
@jonathanprozzi just making sure I understand correctly. you want to show a particular member their overall pctg contribution to the entire guild's contributions to visualize how big their slice of contributions are?

would you need the contribution fragment then? or e.g. just the # of contributions for the dao and # of contributions for the user?

jonathanprozzi:
> @jonathanprozzi just making sure I understand correctly. you want to show a particular member their overall pctg contribution to the entire guild's contributions to visualize how big their slice of contributions are?
> 
> would you need the contribution fragment then? or e.g. just the # of contributions for the dao and # of contributions for the user?

Good questions - apologies for not being clear. That is the usecase -- It would be for this graph:

image

It's a pie showing each member's contribution % out of all contributions for the date range. Definitely right in that we wouldnt need the full Fragment -- just the total # for the member and for the DAO which you're returning with the count already.

Flip-Liquid:
@jonathanprozzi also, would you need the user's display name returned along with the data? or is that already available on the FE?

alexkeating (approved):
Should be good after merge conflicts are resolved

jonathanprozzi:
> 

We have the current user's display name -- we'll want the member names from each DAO if possible to include in the returned data

jonathanprozzi:
Awesome! Once we get this merged in I'll wrap up the Contribution Count ticket too

PR #250, opened 2022-10-13
Add Contribution count Hook

amrro:
## Linear Ticket
[PRO-519 - Remove getUserContributionCount](https://linear.app/govrn/issue/PRO-519/remove-getusercontributioncount)

## Description
Add `useContributionCount` query hook. 

### Changes

- The `startDate` & `endDate` represents the whole day instead of just the date. 
    - `startDate`: the start of day of the current day. ex: `"2021-10-17T22:00:00.000Z"` 
    - `endDate`: the end of day of the current date. ex: `"2022-10-17T21:59:59.999Z"`

PR #263, opened 2022-10-24
nullable userid contribution count

Flip-Liquid:
## Linear Ticket

[Pro-529](https://linear.app/govrn/issue/PRO-529/modify-the-getcontributioncountbydateforuserinrange-to-support)

## Description

Allow a nullable user id to return all contributions for a guild in a timeframe

## Screencaptures


Flip-Liquid:
@jonathanprozzi

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -545,18 +545,30 @@ export class ContributionCustomResolver {
...
+        guildWhere = Prisma.sql`(gc."guild_id" in (${Prisma.join(guildIds)}))`;
+      }
+    }
+
+    let userWhere = Prisma.sql`
+      gc."user_id" = ${userId} OR gc."user_id" is null
+    `; 
+    if (userId == null) {
+      userWhere = Prisma.sql`
+        ${userWhere} OR TRUE
```
alexkeating, line 570: My SQL syntax may be rusty, but shouldn't this be `gc."user_id" = ${userId} OR gc."user_id" is null OR  gc."user_id" is true`? Does OR with IS work?

Flip-Liquid:
@alexkeating @jonathanprozzi looks like this might not be needed after we extend #262 to allow for user id to be supplied. so we won't need this for the pie chart. do we still need this flexibility to supply null user id to get the total number of contributions for a dao grouped by date?

jonathanprozzi:
Once we have #262 in and wire it up with the components we can see if we still want to do this. It was an alternative approach -- we can evaluate if the added flexibility is worth it but I think for now we can focus on getting #262 in and tested!

Thanks @Flip-Liquid !

apps/protocol-api/src/prisma/resolvers/Contribution.ts
Flip-Liquid: the intent here is to have the user clause evaluate to TRUE for all records if the `userId` is not supplied. 

I think the SQL query evaluates to `gc."user_id" = ${userId} OR gc."user_id" is null OR TRUE`, right? The last `OR TRUE` brings the disjunction high for every record

alexkeating:
262 is merged let me know if we still need this

Flip-Liquid:
Let's drop this unless there's a clear use case @alexkeating @jonathanprozzi

jonathanprozzi:
@Flip-Liquid  @alexkeating 

In working to integrate the last chart I realize that we may need this afterall. The bar chart / heat map (PRO-477) needs the data in the shape we get back from this resolver (with the contribution counts per day).

I pulled in these changes to test and by having the `user_id` be nullable we're able to get all the contribution data back per guild.

It would be helpful to return back the `contribution_id` as well so we can get specific Contribution data per the day based on the `id` and our `useContributionGet` -- this would allow us to surface a lot more relevant information similar to how GitHub does.

My SQL chops need work so I couldn't get this to return properly, but I see that we have the `c.id` in the raw SQL. If we could add that to the return that would be helpful!

Flip-Liquid:
@jonathanprozzi @alexkeating fixed another error introduced in #250 which marked all contributions as unassigned

jonathanprozzi (commented):
Seems like the latest change maybe altered something in the query -- seeing this on the User Dashboard:
```
Please try again: Error: Invalid `prisma.$queryRaw()` invocation: Raw query failed. Code: `42P01`. Message: `missing FROM-clause entry for table "g"`: {"response":{"errors":[{"message":"\nInvalid `prisma.$queryRaw()` invocation:\n\n\n Raw query failed. Code: `42P01`. Message: `missing FROM-clause entry for table \"g\"`","locations":[{"line":2,"column":3}],"path":["result"]}],"data":null,"status":500,"headers":{"map":{"content-length":"228","content-type":"application/json; charset=utf-8"}}},"request":{"query":"query getContributionCountByDateForUserInRange($where: GetUserContributionCountInput!) {\n result: getContributionCountByDateForUserInRange(where: $where) {\n count\n date\n guild_id\n name\n }\n}\n","variables":{"where":{"id":5,"startDate":"2021-11-03T04:00:00.000Z","endDate":"2022-11-03T03:59:59.999Z","guildIds":[],"excludeUnassigned":false}}}}
```

Also seeing this error returned in GraphiQL:

```
{
  "errors": [
    {
      "message": "\nInvalid `prisma.$queryRaw()` invocation:\n\n\n  Raw query failed. Code: `42P01`. Message: `missing FROM-clause entry for table \"g\"`",
      "locations": [
        {
          "line": 14,
          "column": 3
        }
      ],
      "path": [
        "getContributionCountByDateForUserInRange"
      ]
    }
  ],
  "data": null
}```

Flip-Liquid:
@jonathanprozzi ok, testing on my local will send a quick iteration

jonathanprozzi:
Working on my end now!

jonathanprozzi (approved):
🔥

PR #265, opened 2022-10-24
[Contract Sync Job] Support Contribution upsert

amrro:
## Linear Ticket
[PRO-535 - Verify contract sync job is working](https://linear.app/govrn/issue/PRO-535/verify-contract-sync-job-is-working)

## Description
To avoid duplicates, contribution will be upserted based on: `chain_id` & `on_chain_id`.


apps/contract-sync-job/src/db.ts
```diff
@@ -63,11 +79,71 @@ export async function getContribution(
...
+        connect: { name: 'minted' },
+      },
+      on_chain_id: contribution.on_chain_id,
+      chain: { connect: { chain_id: `${contribution.chain_id}` } },
+    },
+    update: {
+      name: { set: contribution.name },
+      on_chain_id: { set: contribution.on_chain_id },
+      // proof: { set: contribution.proof },
+      // details: { set: contribution.details },
```
alexkeating, line 112: Why comment these out?

apps/contract-sync-job/src/main.ts
```diff
@@ -1,31 +1,29 @@
...
 const JOB_NAME = 'contract-sync-job';
 
 const CONTRACT_ADDRESS = process.env.CONTRACT_ADDRESS;
 const CHAIN_URL = process.env.CHAIN_URL;
+const CHAIN_ID = 100;
 
 const networkConfig: NetworkConfig = {
   address: CONTRACT_ADDRESS,
-  chainId: 100,
+  chainId: CHAIN_ID,
```
alexkeating, line 26: Nice

apps/contract-sync-job/src/db.ts
amrro: I fixed it: 56aca17cb39c0b83874bd426ab8618b83dbb834b

Prisma accepts only `string | null` values, but it crashes when the value is `undefined`.

alexkeating:
> @alexkeating Is it feasible include `tx_hash` to `GovrnGraphClient.listContributions()`'s return values?

Yea, that is feasible and fine

libs/govrn-subgraph-client/src/lib/subgraph/queries/queries.graphql
```diff
@@ -51,26 +55,3 @@ query listContributions(
     ...contributionFields
   }
 }
-
-fragment guildFields on Guild {
```
Flip-Liquid, line 55: is this related to the PR?

apps/protocol-api/src/main.ts
```diff
@@ -153,6 +155,13 @@ const permissions = shield(
       name: or(isAuthenticated, hasToken),
       users: or(isAuthenticated, hasToken),
     },
+    Chain: {
+      id: hasToken,
```
Flip-Liquid, line 159: why no isAuthenticated?

libs/govrn-subgraph-client/src/lib/subgraph/queries/queries.graphql
alexkeating: Yea, we need to add fields to the subgraph and this prevented us from adding those fields

apps/protocol-api/src/main.ts
alexkeating: Because it is not used by the frontend, best to only add permissions when needed.

PR #267, opened 2022-10-26
feature/dao-dashboard-contributions-by-type (PRO-498)

jonathanprozzi:
## Linear Ticket

- [PRO-498](https://linear.app/govrn/issue/PRO-498/[pie]-contributions-by-type)
- [PRO-563](https://linear.app/govrn/issue/PRO-563/create-usecontributionactivitytype-hook)

## Description

- Adds a `useContributionActivityType` hook that calls the new endpoint
- Adds the _Contributions by Type_ pie chart 

Data flows from DaoDashboardShell (passes in the DAO id from the query param) -> ContributionPieShell (hook call for the data) -> ContributionTypesPie
- Allows for the error / fetching states (and spinner UI) to be at the chart level instead of at the page level
- We can adjust this in the future based on feedback and any error boundaries we choose to implement
- Chart is responsive and some of the values are different for mobile -- we'll tweak the sizes once everything is in place
- There are hardcoded values for the `startDate` and `endDate` args -- this'll be passed in via the Date Picker but added this for now for testing

## Screencaptures

https://user-images.githubusercontent.com/9438776/197932117-f8d520ea-72fe-4544-9067-a869b080c7c1.mp4

I'll adjust the sizing a bit more in 550 — will have a better idea once all charts are in


apps/protocol-frontend/src/components/ContributionTypesPie.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const brandColorMap = [
+  '#ffb4e2',
+  '#fb84ce',
+  '#f854ba',
+  '#f526a6',
+  '#db0f8d',
+  '#ab076d',
+  '#76024e',
+];
```
alexkeating, line 25: Do we have a central place to put our css color palette? Are these the came on the bar chart

jonathanprozzi: Ah, I was thinking about this -- we have these in our `theme.primary` object in the theme, but this was a convenient mapping of those values in a format we need for these graphs. I'm thinking we could add this to the theme somewhere and export as a constant from there incase we want to update it -- that way it's still in 1 place.

Alternatively we could map from the theme palette itself, that way at least that's the source of truth. I think I may have had that originally for the heat map and removed it. Great call out though @alexkeating -- we should do 1 or the other so we can keep this easily updatable

PR #268, opened 2022-10-27
expose getContributionCount query, add hook

Flip-Liquid:
## Linear Ticket

[Pro-528](https://linear.app/govrn/issue/PRO-528/create-a-usedaocontributioncount-hook)

## Description

Add hook for useContributionCount

## Screencaptures


apps/protocol-frontend/src/hooks/useDaoContributionCount.ts
```diff
@@ -0,0 +1,27 @@
...
+const useDaoContributionCount = (args: {
+  startDate: Date;
+  endDate: Date;
+  guildIds?: number[] | null;
+  excludeUnassigned?: boolean;
+}) => {
+  const { userData, govrnProtocol: govrn } = useUser();
+
+  const { isLoading, isFetching, isError, error, data } = useQuery(
+    ['contributionGetCount', args],
```
alexkeating, line 13: This should probably be daoContributionGetCount now

apps/protocol-frontend/src/hooks/useDaoContributionCount.ts
```diff
@@ -0,0 +1,27 @@
+import { useQuery } from '@tanstack/react-query';
+import { useUser } from '../contexts/UserContext';
+
+const useDaoContributionCount = (args: {
+  startDate: Date;
+  endDate: Date;
+  guildIds?: number[] | null;
```
alexkeating, line 7: This should be guildId not guildIds

apps/protocol-frontend/src/hooks/useDaoContributionCount.ts
```diff
@@ -0,0 +1,27 @@
...
+  excludeUnassigned?: boolean;
+}) => {
+  const { userData, govrnProtocol: govrn } = useUser();
+
+  const { isLoading, isFetching, isError, error, data } = useQuery(
+    ['contributionGetCount', args],
+    async () => {
+      return await govrn.custom.getDaoContributionCount({
+        userId: userData?.id,
+        ...args,
```
alexkeating, line 17: Replace with guildId key

apps/protocol-frontend/src/hooks/useDaoContributionCount.ts
jonathanprozzi: also won't be an array of numbers -- should likely just be a `number`

jonathanprozzi (commented):
Small suggestions (also agree with Keating's) but this looks good. Thanks for the quick change after yesterday's sync!

apps/protocol-frontend/src/hooks/useDaoContributionCount.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { isLoading, isFetching, isError, error, data } = useQuery(
+    ['contributionGetCount', args],
+    async () => {
+      if (!userData?.id) {
+        console.error('getUserContributionsCount has no userData.id');
+        return [];
+      }
+
+      return await govrn.custom.getContributionCount({
+        userId: userData?.id,
```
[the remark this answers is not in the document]
Flip-Liquid:
```
{
  getDaoContributionCount(where: {
    guildId: 2, startDate: "2021-12-21 00:00:00.000", endDate: "2022-12-21 00:00:00.000"
  }) 
}
```

jonathanprozzi (commented):
@Flip-Liquid I think that the types for `useDaoContributionCount` and `useContributionCount` may have been reversed -- `useContributionCount` is meant to take in the array of `guildIds` since it's used in the UserDashboard

Flip-Liquid:
they absolutely were flipped 🤦 @jonathanprozzi . should be fixed now. tested the query again, and it looked like it was working from graphiql

PR #269, opened 2022-10-27
feature/dao-dashboard-timeframe-setter (PRO-481)

jonathanprozzi:
## Linear Ticket

- [PRO-481](https://linear.app/govrn/issue/PRO-481/timeframe-setter)

## Description

Adds a first iteration for the timeframe setter to the DAO Dashboard
- Supports the same pre-defined values as the User Dashboard accessed via the dropdown (we can tweak these based on feedback)
- Supports a _Custom_ date range -- choosing 'Custom' in the select opens the date picker where the user can select a range
  - The datepicker is only visible if Custom is selected so that it doesn't add to the UI unnecessarily
- Passes this to the components that accept date ranges

Thoughts:
- Design is a bit different than the original wireframe. I had an iteration where the Custom date range value shows up in the menu (similar to the wireframe) but this caused a lot of side effects
- Very much open to altering the design (and the pre-defined date ranges) based on feedback, but also wanted to put up the first iteration for testing purposes


libs/protocol-ui/src/components/molecules/ControlledSelect/ControlledSelect.tsx
```diff
@@ -24,9 +24,9 @@ export interface ControlledSelectProps {
   isMulti?: boolean;
   isClearable?: boolean;
   onChange?: (option: { value: number; label: string }) => void;
-  isDisabled?: boolean;
   variant?: 'outline' | 'filled';
   value?: any;
```
jonathanprozzi, line 28: This is lingering from our type fixes -- I removed about 30 warnings and cleaned the UI package up considerably, but this is on my list

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -1,24 +1,110 @@
...
 }
-
-const startDate = new Date('7/1/2022'); // temporary hardcoded value while working on the datepicker component
-const endDate = new Date('8/1/2022'); // temporary hardcoded value while working on the datepicker component
+const TODAY_DATE = new Date();
 
 const DaoDashboardShell = ({ daoName, daoId }: DaoDashboardShellProps) => {
+  const [showCustomDatePicker, setShowCustomDatePicker] = useState(false);
+
+  const [startDate, setStartDate] = useState<Date>(new Date(TODAY_DATE));
```
alexkeating, line 18: Isn't TODAY_DATE a Date already?

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -1,24 +1,110 @@
...
+                  setEndDate(end);
+                }}
+                startDate={startDate}
+                endDate={endDate}
+                maxDate={new Date(TODAY_DATE)}
+              />
+            </Box>
+            <ControlledSelect
+              isSearchable={false}
+              defaultValue={dateRangeOptions.find(date => date.value === 4)}
```
alexkeating, line 96: Can we use a const for 4? something like Month I guess? It could make the code more readable

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export interface ControlledDatePickerProps extends ReactDatePickerProps {
+  onChange: (
+    date: Date[] | Date | [Date | null, Date | null] | null,
+    event: SyntheticEvent<Date, Event> | undefined,
+  ) => void;
+  startDate: Date | null | undefined;
+  endDate: Date | null | undefined;
+  maxDate?: Date | null | undefined;
+}
```
alexkeating, line 15: Is this necessary? Couldn't we use ReactDatePickerProps? Some of these look like what is already in there.

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePickerButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ref?: HTMLInputElement;
+  buttonDisabled?: boolean;
+};
+
+export type CustomDatePickerButton = ComponentWithAs<
+  'button',
+  ControlledDatePickerButtonProps
+> & {
+  Button: typeof Button;
+};
```
alexkeating, line 22: Is this used anywhere else?

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePickerButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  forwardRef,
+  ComponentWithAs,
+} from '@chakra-ui/react';
+import { HiOutlineCalendar } from 'react-icons/hi';
+
+export type ControlledDatePickerButtonProps = ButtonProps & {
+  value?: React.ReactNode;
+  onClick?: () => void;
+  ref?: HTMLInputElement;
+  buttonDisabled?: boolean;
```
alexkeating, line 14: Isn't disabled a prop in ButtonProps?

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -1,24 +1,110 @@
...
+
+  const dateRangeOptions = [
+    {
+      value: 'Custom',
+      label: 'Custom',
+    },
+    { value: 1, label: 'Last Week' },
+    { value: 4, label: 'Last Month' },
+    { value: 12, label: 'Last Quarter' },
+    { value: 52, label: 'Last Year' },
```
alexkeating, line 35: It may be good to move these number to name consts. We could probably reuse them in the UserDashboard to make it clearer what is going on in the code.

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -1,24 +1,110 @@
...
+            width="100%"
+            flexBasis={{ base: '100%', lg: '60%' }}
+            flexGrow="1"
+            gap={{ base: 0, lg: 2 }}
+          >
+            <Box visibility={showCustomDatePicker ? 'inherit' : 'hidden'}>
+              <ControlledDatePicker
+                selected={startDate}
+                onChange={dates => {
+                  const datesArray = dates as Date[];
```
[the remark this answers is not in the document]
jonathanprozzi: Yeah -- I had some difficulty with this. The onChange is an event, but since this is a date range from the date picker the values for start and end are coming in as an array of Dates. I tried extending the date type to allow for this, but had some conflicts coming from DatePicker's types. This doesn't feel quite right to me either, but the `dates` is the synthetic event and then the `datesArray` is an array of date objects.

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
```diff
@@ -1,24 +1,110 @@
...
+                  const [start, end] = datesArray;
+                  setStartDate(start);
+                  setEndDate(end);
+                }}
+                startDate={startDate}
+                endDate={endDate}
+                maxDate={new Date(TODAY_DATE)}
```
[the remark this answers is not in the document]
jonathanprozzi: Good catch, will clean that up

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePickerButton.tsx
jonathanprozzi: It's used in both the custom buttons for the Date Picker -- without this extension of the `ComponentWithAs` Chakra has trouble picking up the types

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePickerButton.tsx
jonathanprozzi: Yeah -- I had an issue with a conflict when passing it in, but will revisit that and try again since this is a bit strange

jonathanprozzi: Also we can remove this completely -- I opted for setting the visibility to hidden instead of disabling the button. I disabled it previously when Custom wasn't selected. Going to remove

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
jonathanprozzi: Great idea, will help clean things up and also make it so we can more directly update this if we want to change our pre-defined date values

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePicker.tsx
jonathanprozzi: Ah, yeah. That's a good point. Let me try that -- I had some issues with this originally but I think it may have been from something else. If these are unnecessary I'll update our other DatePicker as well 👍

jonathanprozzi: Works since spreading props and extending -- changed the DatePicker to use this pattern

apps/protocol-frontend/src/components/DaoDashboardShell.tsx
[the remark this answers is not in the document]
jonathanprozzi: Definitely feels off -- I can revisit this but I had some trouble with this due to then mismatching with the React DatePicker types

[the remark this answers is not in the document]
jonathanprozzi: Digging into it a bit more -- without casting the dates as the Date[] there is an issue with the type not having an iterator. I've fixed this before by spreading, but that doesn't seem to be working here.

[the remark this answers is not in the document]
alexkeating: hm, yea, I think this is fine after looking at the DatePicker props I think they have a special signature like you were saying for the onChange prop..

libs/protocol-ui/src/components/molecules/ControlledDatePicker/ControlledDatePickerButton.tsx
alexkeating: Ah okay my bad

alexkeating (approved):

PR #273, opened 2022-10-28
Ipfs update meta (#272)

alexkeating:
## Linear Ticket

[PRO-534 - Add activity type id/name and contribution id to details URI](https://linear.app/govrn/issue/PRO-534/add-activity-type-idname-and-contribution-id-to-details-uri)

## Description
Update IPFS storage to this format: 
```ts
{
  version: number, // version 1
  name: string,
  details: string,
  proof: string,
  activityName: string,
  image: string, // keep blank for now
  govrn: { // Contain govrn specific data
    id: number, // Contribution id
    activityTypeId: numer;
  }
}
```


libs/protocol-client/src/lib/utils/ipfs.ts
```diff
@@ -0,0 +1,73 @@
+import { Buffer } from 'buffer';
+import { create, IPFSHTTPClient, CID } from 'ipfs-http-client';
+import { ContributionV1SchemaCid } from './constants';
+
+export type MintedContributionSchemaV1 = {
+  schema: `ipfs://Qmc3fKseQ1JyiJiPsmazx23A6twy6YCkTfqDfHUKZDL5xq`;
```
amrro, line 6: `schema` is omitted for all usages, and sometimes replaced like here.

apps/protocol-frontend/src/libs/ipfs.ts
```diff
@@ -1,64 +1,40 @@
...
   });
 };
+
+// 1. some clean up
+// 2. Add comand to generate schema uri
+// 3. Test schema matches type
+// 4. test store contribution metadata
+// 5. test mint hook to make sure proper uri is passed in
+// 6. Test uploadFileIPFS as well
+// 7. Move schema to protocol-client
```
amrro, line 40: Is this necessary?

libs/protocol-client/src/lib/utils/ipfs.ts
```diff
@@ -0,0 +1,73 @@
...
+  activityName: string;
+  image?: string; // keep blank for now
+  govrn: {
+    /* Contribution ID */
+    id: number;
+    activityTypeId: number;
+  };
+};
+
+export class IPFS {
```
amrro, line 19: Nice 👍🏻.

amrro (changes requested):
Some minor changes. 

Other than that great work. Small tests & IPFS class is great addition.

apps/protocol-frontend/src/libs/ipfs.ts
alexkeating: Nice thanks will remove

libs/protocol-client/src/lib/utils/ipfs.ts
```diff
@@ -0,0 +1,73 @@
...
+  public async storeContributionMetadata(
+    content: Omit<MintedContributionSchemaV1, 'schema'>,
+  ) {
+    const contentWithVersion = {
+      ...content,
+      schema: ContributionV1SchemaCid,
+      version: this.VERSION,
+    };
+    const cid = await this.client.add(JSON.stringify(contentWithVersion), {
+      cidVersion: 1,
```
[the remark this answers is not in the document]
alexkeating: Cid version is something that is ipfs specific, and not related to our schema which we define on Govrn.

libs/protocol-client/src/lib/utils/ipfs.ts
alexkeating: This is a type definition here. It is saying the type of the schema is that string. The reason we omit schema sometimes is we only need to define it at the lowest call level which would be at the line below.

https://github.com/Govrn-HQ/govrn-monorepo/pull/273/files#diff-44778fada2dbebb5a7858bcb7ae0fdf3a7e3d05e69bd44e8b4f3fdb70fd6ec93R32

libs/protocol-client/src/lib/utils/ipfs.ts
```diff
@@ -0,0 +1,71 @@
+import { Buffer } from 'buffer';
+import { create, IPFSHTTPClient, CID } from 'ipfs-http-client';
+import { ContributionV1SchemaCid } from './constants';
+
+export type MintedContributionSchemaV1 = {
+  schema: `ipfs://Qmc3fKseQ1JyiJiPsmazx23A6twy6YCkTfqDfHUKZDL5xq`;
```
jonathanprozzi, line 6: For my own understanding -- this schema value is the text that is the `ipfs://` with the `ContributionV1SchemaCid` appended after the // 

Would we want to use the constant there? thinking of if we update the value in the constants but not here if that would cause issues

libs/protocol-client/src/lib/utils/ipfs.ts
```diff
@@ -0,0 +1,71 @@
...
+  name: string;
+  details: string;
+  proof: string;
+  activityName: string;
+  image?: string; // keep blank for now
```
jonathanprozzi, line 11: Thinking that this could potentially be a separate field for an image to allow for uploads of images since our proof field is more of a catchall?

libs/protocol-client/src/tests/utils/ipfs.test.ts
```diff
@@ -0,0 +1,26 @@
+import { describe, expect, test } from '@jest/globals';
+import Ajv from 'ajv';
+import v1Schema from '../../schemas/contributionMetadataV1.json';
+import { MintedContributionSchemaV1 } from '../../lib/utils/ipfs';
+import { ContributionV1SchemaCid } from '../../lib/utils/constants';
+
+describe('IPFS test', () => {
+  test('test schema matches type', () => {
+    const ajv = new Ajv(); // options can be passed, e.g. {allErrors: true}
+    const validate = ajv.compile(v1Schema);
```
jonathanprozzi, line 10: Ah that's so cool!

apps/protocol-frontend/src/hooks/useContributionMetadataStore.ts
```diff
@@ -0,0 +1,27 @@
+import { useMutation } from '@tanstack/react-query';
+import { INFURA_PROJECT_ID, INFURA_PROJECT_SECRET } from '../utils/constants';
+import { MintedContributionSchemaV1, IPFS } from '@govrn/protocol-client';
+
+const useContributionMetadataStore = () => {
+  const { mutateAsync, isLoading, isError, isSuccess } = useMutation(
+    ['ContributionMetadataStore'],
```
jonathanprozzi, line 7: I don't think we need to pass a key into the mutations -- what I've seen is that it's more important for the queries due to the invalidation by key. But this'll help with any debugging we want to do with the React Query dev tools. 

If we find that it's helpful we can add keys to our other mutations too

apps/protocol-frontend/src/hooks/useContributionMetadataStore.ts
```diff
@@ -0,0 +1,27 @@
+import { useMutation } from '@tanstack/react-query';
+import { INFURA_PROJECT_ID, INFURA_PROJECT_SECRET } from '../utils/constants';
+import { MintedContributionSchemaV1, IPFS } from '@govrn/protocol-client';
+
+const useContributionMetadataStore = () => {
```
jonathanprozzi, line 5: Love that we're using `useMutation` for these IPFS updates -- it's so flexible

apps/protocol-frontend/src/hooks/useContributionMetadataStore.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { mutateAsync, isLoading, isError, isSuccess } = useMutation(
+    ['ContributionMetadataStore'],
+    async (data: Omit<MintedContributionSchemaV1, 'schema'>) => {
```
jonathanprozzi, line 8: Nice, so when/if we update that Schema we'll have the type safety here

apps/protocol-frontend/src/hooks/useContributionMetadataStore.ts
```diff
@@ -0,0 +1,27 @@
...
+      const ipfs = new IPFS(INFURA_PROJECT_ID, INFURA_PROJECT_SECRET);
+      return await ipfs.storeContributionMetadata({
+        ...data,
+      });
+    },
+    {
+      onSuccess: (data, ipfsURI) => {
+        return ipfsURI;
```
jonathanprozzi, line 16: We probably don't want to display a toast here do we? I'm thinking probably not since this is called during other actions such as the bulk mint. 

Could be too many Toasts, but nice that we'd have that option if we wanted to use this for something else (or add the toast regardless, but feeling like having a Toast for this and another on Bulk Mint success is too much)

apps/protocol-frontend/src/hooks/useContributionMint.ts
```diff
@@ -16,6 +17,8 @@ const useContributionMint = () => {
   const { data: signer } = useSigner();
   const { chain } = useNetwork();
   const { userData, govrnProtocol: govrn } = useUser();
+  const { mutateAsync: storeContributionMetadata } =
```
jonathanprozzi, line 20: Nice!

jonathanprozzi (commented):
Small questions mostly for my own sake! Looks great, going to test some actual minting too before fully approving

jonathanprozzi (approved):
bulk Mint worked on Goerli when testing -- awesome work! 🔥

libs/protocol-client/src/lib/utils/ipfs.ts
amrro: I tried that. But TS doesn't allow you to use value as a type.

libs/protocol-client/src/lib/utils/ipfs.ts
alexkeating: Image is more of like an NFT image to represent the contribution.

libs/protocol-client/src/tests/utils/ipfs.test.ts
alexkeating: yea, if there is a mismatch we probably need a new schema

apps/protocol-frontend/src/hooks/useContributionMetadataStore.ts
alexkeating: yea

apps/protocol-frontend/src/hooks/useContributionMetadataStore.ts
alexkeating: Good call I can add the toast as a parameter

jonathanprozzi:
Nice addition with the successToast as a variable. That's a great pattern we can follow in the future where needed

PR #279, opened 2022-10-31
Contract sync batching

amrro:
## Linear Ticket

[PRO-536](https://linear.app/govrn/issue/PRO-536/make-sure-job-processes-contributions-concurrently)

## Description
- Batch requests, especially for db.
- Change batching method to use [promise pool](https://github.com/Govrn-HQ/govrn-monorepo/pull/279).


apps/contract-sync-job/src/db.ts
```diff
@@ -74,9 +74,10 @@ export async function getContribution(data: { tokenId: number }) {
     ).result;
 
     if (contrs.length != 0) return contrs[0].id;
-    throw new Error('Contribution must exist for this attestation!');
+    throw new Error('Contribution is missing.');
```
alexkeating, line 77: This error message should probably be more clear. I think the original is more specific then this change.

alexkeating (commented):
Looks great! One comment then we can merge

jonathanprozzi (commented):
This is awesome. Agree with Keating about the error message but this is great 🔥🔥

PR #280, opened 2022-10-31
dao contributions by user

Flip-Liquid:
## Linear Ticket

[Pro-568](https://linear.app/govrn/issue/PRO-568/create-endpoint-for-dao-contributions-by-member-in-date-range)

## Description

add query and client method for getting dao contributions by user

## Screencaptures

### DIFF
--- a/apps/protocol-api/src/prisma/resolvers/Contribution.ts
+++ b/apps/protocol-api/src/prisma/resolvers/Contribution.ts
@@ -280,7 +280,7 @@ export class ContributionCountByUser {
   @TypeGraphQL.Field(_type => String, {
     nullable: true
   })
-  user_name?: string;
+  display_name?: string;
 }
 
 @TypeGraphQL.ObjectType('TotalContributionCount', { isAbstract: true })
@@ -727,7 +727,7 @@ export class ContributionCustomResolver {
     const result = await prisma.$queryRaw<ContributionCountByUser>`
       SELECT  count(gc.id) as count,
               u.id as "user_id",
-              u.display_name as "user_name"
+              u.display_name as "display_name"
       FROM
           "GuildContribution" as gc 
           LEFT JOIN "Contribution" as c
--- a/apps/protocol-api/src/prisma/resolvers/Contribution.ts
+++ b/apps/protocol-api/src/prisma/resolvers/Contribution.ts
@@ -267,6 +267,22 @@ export class ContributionCountByDate {
   name: string;
 }
 
+@TypeGraphQL.ObjectType('ContributionCountByUser', { isAbstract: true })
+export class ContributionCountByUser {
+  @TypeGraphQL.Field(_type => Number)
+  count: number;
+
+  @TypeGraphQL.Field(_type => Number, {
+    nullable: true
+  })
+  user_id?: number;
+
+  @TypeGraphQL.Field(_type => String, {
+    nullable: true
+  })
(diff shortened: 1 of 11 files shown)


libs/protocol-client/src/lib/graphql/queries.graphql
```diff
@@ -451,6 +451,16 @@ query getDaoContributionCount(
   result: getDaoContributionCount(where: $where) 
 }
 
+query getDaoContributionCountByUser(
+  $where: GetContributionInput!
+) {
+  result: getDaoContributionCountByUser(where: $where) {
+    count
+    user_name
```
jonathanprozzi, line 459: Nice to have `user_name` included on the return values -- perfect for the pie chart 🔥

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -696,4 +712,33 @@ export class ContributionCustomResolver {
...
+    @TypeGraphQL.Args() args: GetContributionArgs,
+  ) {
+    const start = args.where.startDate;
+    const end = args.where.endDate;
+    const guildId = args.where.guildId;
+
+    const result = await prisma.$queryRaw<ContributionCountByUser>`
+      SELECT  count(gc.id) as count,
+              u.id as "user_id",
+              u.display_name as "user_name"
```
jonathanprozzi, line 730: Wondering if we may want to standardize when/how we use `display_namename` instead of `name` -- our profile call to `user.create` sets the username (which I think is the `name` field in the user table) 

Does Kevin Malone write to `display_name`? Curious mainly so we're on the same page -- this looks great, that's my only question!

jonathanprozzi (approved):
Approved! Question in comment about `display_name` vs `name` vs `user_name` and which we should be using -- see the `u.display_name` so just wondering since we set `username` / `user.name`. Which does Kevin Malone set? We should all be sure to align!

Awesome job 🔥

apps/protocol-api/src/prisma/resolvers/Contribution.ts
Flip-Liquid: Kevin updates both name and display_name in the User table. but I agree with you that it's not immediately clear why there are two separate fields for this. @alexkeating do you have an opinion on this?

alexkeating: Yes, we should probably consolidate. I think originally display name was meant to be discord name and the name was like there user name but I think we added display name to the discord table making the distinction irrelevant in user

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -267,6 +267,22 @@ export class ContributionCountByDate {
...
+
+  @TypeGraphQL.Field(_type => Number, {
+    nullable: true
+  })
+  user_id?: number;
+
+  @TypeGraphQL.Field(_type => String, {
+    nullable: true
+  })
+  user_name?: string;
```
alexkeating, line 283: probably should rename this to display_name to avoid any confusion with name and username.

alexkeating (commented):
Looks good, I think one field needs to be renamed

Flip-Liquid:
@alexkeating @jonathanprozzi sent a quick iteration with the suggested renaming

alexkeating (approved):

PR #282, opened 2022-11-02
Add Bulk Attest to SDK

amrro:
## Linear Ticket

[PRO-440](https://linear.app/govrn/issue/PRO-440/create-and-use-bulkattest)

## Description

Add `bulkAttest` method to SDK.

### DIFF
--- a/libs/protocol-client/src/lib/client/contribution.ts
+++ b/libs/protocol-client/src/lib/client/contribution.ts
@@ -267,25 +267,27 @@ export class Contribution extends BaseClient {
     const transaction = await contract.bulkAttest(args);
     const transactionReceipt = await transaction.wait();
 
-    const logs = transactionReceipt.logs.map((l, idx) => {
+    const logs: {
+      index: number;
+      contributionId: number;
+      confidence: number;
+    }[] = [];
+    for (const [index, l] of transactionReceipt.logs.entries()) {
       const log = contract.govrn.interface.parseLog(l);
-      return log.name === 'Attest'
-        ? {
-            index: idx,
-            contributionId: (log.args['contribution'] as BigNumber).toNumber(),
-            confidence: log.args['confidence'],
-          }
-        : /* Technically, this will never happen.
-            Attestations will either be all successful or all failed. */
-          null;
-    });
+      if (log.name === 'Attest') {
+        logs.push({
+          index,
+          contributionId: (log.args['contribution'] as BigNumber).toNumber(),
+          confidence: log.args['confidence'],
+        });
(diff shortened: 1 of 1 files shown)


libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -241,6 +241,58 @@ export class Contribution extends BaseClient {
...
 
+  public async bulkAttest(
+    networkConfig: NetworkConfig,
+    signer: ethers.Signer,
+    chainId: number,
+    userId: number,
+    data: {
+      contribution: number;
+      confidenceId: number;
+      confidenceName: string;
```
amrro, line 252: I've to pass confidence's name & id because it's required for create on chain attestation.

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -241,6 +241,58 @@ export class Contribution extends BaseClient {
...
+      return log.name === 'Attest'
+        ? {
+            index: idx,
+            contributionId: (log.args['contribution'] as BigNumber).toNumber(),
+            confidence: log.args['confidence'],
+          }
+        : /* Technically, this will never happen.
+            Attestations will either be all successful or all failed. */
+          null;
+    });
```
alexkeating, line 281: A for loop would probably be better. We don't want to have branches of logic that never occur.

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -241,6 +241,58 @@ export class Contribution extends BaseClient {
...
+    const contract = new GovrnContract(networkConfig, signer);
+    const transaction = await contract.bulkAttest(args);
+    const transactionReceipt = await transaction.wait();
+
+    const logs = transactionReceipt.logs.map((l, idx) => {
+      const log = contract.govrn.interface.parseLog(l);
+      return log.name === 'Attest'
+        ? {
+            index: idx,
+            contributionId: (log.args['contribution'] as BigNumber).toNumber(),
```
alexkeating, line 275: What type is this? Why does it need to be explicitly set as a BigNumber?

amrro: All `log.args`'s fields are `any` type.

amrro: Furthermore, this is the type `args`: 
```ts
export interface Result extends ReadonlyArray<any> {
    readonly [key: string]: any;
}
```

alexkeating (approved):

PR #283, opened 2022-11-03
Add Bulk Attest to UI

amrro:
## Linear Ticket
[PRO-440](https://linear.app/govrn/issue/PRO-440/create-and-use-bulkattest)
____
Related #282

### DIFF
--- a/apps/protocol-frontend/src/components/BulkAttestationModal.tsx
+++ b/apps/protocol-frontend/src/components/BulkAttestationModal.tsx
@@ -36,7 +36,6 @@ const BulkAttestationModal = ({ contributions }: BulkAttestationModalProps) => {
             onChainId: c.onChainId,
             confidence: 1,
             confidenceName: 'Verified',
-            dateOfSubmission: 0,
           };
         }),
       );
--- a/apps/protocol-frontend/src/components/BulkAttestationModal.tsx
+++ b/apps/protocol-frontend/src/components/BulkAttestationModal.tsx
@@ -13,7 +13,7 @@ import { useUser } from '../contexts/UserContext';
 import { MdCheckCircle } from 'react-icons/all';
 import { AttestationTableType } from '../types/table';
 import pluralize from 'pluralize';
-import useAttestationMint from '../hooks/useAttestationMint';
+import useAttestationBulkMint from '../hooks/useAttestationBulkMint';
 
 interface BulkAttestationModalProps {
   contributions: AttestationTableType[];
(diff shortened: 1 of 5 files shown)


amrro:
Note:
This is error popups with Cypress tests: 
```Something went wrong. Please try again: The method eth_blockNumber is not implemented by the mock provider.```

This is why I had to wrap `bulkAttest` into try...catch and Keating did the same with bulk mint in #248.

**My question: are we wasting resources by repeatedly running these tests**?

File: apps/protocol-frontend/src/components/BulkAttestationModal.tsx [hunk 1]
```diff
@@ -13,26 +13,36 @@ import { useUser } from '../contexts/UserContext';
...
-      try {
-        await mintAttestation(contribution);
-      } catch {
-        return;
-      }
-    });
+  const createAttestationsHandler = async (
+    contributions: AttestationTableType[],
+  ) => {
+    try {
```
alexkeating, line 31: Should we put this try catch in the hook and remove the handler?

apps/protocol-frontend/src/components/BulkAttestationModal.tsx
```diff
@@ -13,26 +13,36 @@ import { useUser } from '../contexts/UserContext';
 import { MdCheckCircle } from 'react-icons/all';
 import { AttestationTableType } from '../types/table';
```
alexkeating, line 14: (random spot to start thread), when bulk attesting shouldn't the attest button become bulk attest?

alexkeating: Modal doesn't close when attesting is done

alexkeating:
> Note: This is error popups with Cypress tests: `Something went wrong. Please try again: The method eth_blockNumber is not implemented by the mock provider.`
> 
> This is why I had to wrap `bulkAttest` into try...catch and Keating did the same with bulk mint in #248.
> 
> **My question: are we wasting resources by repeatedly running these tests**?

I don't think so we have caught several bugs, and it doesn't cost us anything. The reason those errors came about was because some flow did change.

alexkeating (changes requested):

apps/protocol-frontend/src/components/BulkAttestationModal.tsx
amrro: query actually never fails, it display the error gracefully. Also, query supposedly have a error handling already. 
It display the `eth_blocknumber` gracefully.

alexkeating: It will fail if the rpc is down. I don't think onError catches the error instead it executes when an error occur.

alexkeating (approved):

PR #284, opened 2022-11-03
feature/dao-dashboard-contributions-by-date (PRO-477)

jonathanprozzi:
## Linear Ticket

- [PRO-477](https://linear.app/govrn/issue/PRO-477/[bar]-contributions-by-date-chart)

## Description

- Adds in a Calendar Heatmap showing the DAO's Contributions over the past year
- Hovering on the day shows a snapshot/"daily insight" of the Contributions that day

Notes: 

- The Custom Tooltip component is in the same file as the Heatmap. Ideally this would be a separate component, but the `tooltip` prop doesn't want to take in a full component so I can't pass the props in -- it's context sensitive since it is aware of the `day` and `value` coming from the hover event
- We can likely make additional optimizations for the hover feature, but this is a first iteration of us creating a context sensitive feed -- potentially down the road we can open a 'feed' / new page with more detail!
- Positioned the tooltip so that it doesn't push off the screen
- Additional UI adjustments for the whole layout will be in an upcoming PR

## Screencaptures

https://user-images.githubusercontent.com/9438776/199850835-37b5e6b0-2f68-4df1-8414-a6e0e9e3c482.mp4

Daily Contributions tooltip snapshot (desktop):

image

Daily Contributions tooltip snapshot (mobile):

image


jonathanprozzi:
One strange thing I'm noticing while I was testing this PR is that the data shown in the daily snapshot sometimes isn't fully accurate -- I notice that if I go and do a bulk update to attribute them to a DAO they show up, but if I create them all using the 'create more' they won't. I'm not sure this relates to this feature itself, but going to check it out a bit further.

jonathanprozzi:
Updates:
- Resolves issue with the inconsistencies in the data and removed the `formatDate` since it was using localized time and wasn't accurate on the hover

image

apps/protocol-frontend/src/hooks/useContributionCount.ts
```diff
@@ -2,25 +2,22 @@ import { useQuery } from '@tanstack/react-query';
...
-
-      return await govrn.custom.getContributionCountByDateForUserInRange({
-        id: userData?.id,
-        ...args,
-      });
+      return await govrn.custom.getContributionCountByDateForUserInRange(args);
+      // return await govrn.custom.getContributionCountByDateForUserInRange({
+      //   id: userData?.id,
+      //   ...args,
+      // });
```
alexkeating, line 20: This should either be removed or return some value.

apps/protocol-frontend/src/hooks/useContributionCount.ts
```diff
@@ -2,25 +2,22 @@ import { useQuery } from '@tanstack/react-query';
 import { useUser } from '../contexts/UserContext';
 
 const useContributionCountInRange = (args: {
+  id?: number | null | undefined;
```
alexkeating, line 5: undefined is redundant

apps/protocol-frontend/src/components/ContributionsByDateShell.tsx
```diff
@@ -0,0 +1,37 @@
...
+const ContributionByDateShell = ({
+  guildIds,
+}: ContributionByDateShellProps) => {
+  const {
+    isFetching,
+    isLoading,
+    isError,
+    data: contributionsByDate,
+  } = useContributionCountInRange({
+    id: null,
```
alexkeating, line 19: Does null vs undefine matter here?

apps/protocol-frontend/src/components/ContributionsByDateShell.tsx
```diff
@@ -0,0 +1,37 @@
...
+    excludeUnassigned: true,
+  });
+
+  return (
+    <ContributionsByDateChart
+      contributionsCount={contributionsByDate}
+      isFetching={isFetching}
+      isLoading={isLoading}
+      isError={isError}
+      daoId={guildIds[0]}
```
alexkeating, line 32: what if guildIds is an empty array?

apps/protocol-frontend/src/components/ContributionsByDateChart.tsx
```diff
@@ -0,0 +1,200 @@
...
+            width="100%"
+            paddingX={{ base: 2, lg: 0 }}
+          >
+            <ResponsiveTimeRange
+              data={
+                contributionsCountMap
+                  ? contributionsCountMap.filter(
+                      contribution => contribution.value !== 0,
+                    )
+                  : []
```
alexkeating, line 140: Is this check necessary with the check above?

apps/protocol-frontend/src/components/ContributionsByDateChart.tsx
```diff
@@ -0,0 +1,200 @@
...
+  if (isFetching || isLoading) {
+    return <GovrnSpinner />;
+  }
+
+  const CustomTooltipDailyContributions = ({
+    day,
+    value,
+  }: CalendarTooltipProps) => {
+    const { data: dailyContributions } = useContributionList({
+      first: 1000, // TODO: better way to do this -- do we want to have to use infinite list for this? we can also cap it at 10 or something
```
alexkeating, line 61: Since this is in the tooltip I would say only show max 5 and order them by hightest attestation count. We probably shouldn't render 1000+ contributions in a tooltip

alexkeating (changes requested):

apps/protocol-frontend/src/components/ContributionsByDateChart.tsx
jonathanprozzi: Yeah for sure -- I was thinking 5 max is good -- definitely wasn't intending to render 1000. Probably 5 is a reasonable limit and then we can go from there

apps/protocol-frontend/src/components/ContributionsByDateShell.tsx
jonathanprozzi: I added some additional protection here for this, but it's coming from the query param and is set to an array to match the hook structure which is looking for an array of numbers since we use it in the user dashboard.

If someone doesnt enter the param in the url (feature/dao/ ) for example, the page is a 404 anyway so this won't be rendered

apps/protocol-frontend/src/components/ContributionsByDateShell.tsx
alexkeating: Ah okay so it does then was trying to understand if that is how we have it on the backend

apps/protocol-frontend/src/components/ContributionsByDateChart.tsx
alexkeating: I was talking more about the turnery expression. Doesn't line 253 already do this check. We can have the filter without the turnery.

apps/protocol-frontend/src/components/ContributionsByDateShell.tsx
alexkeating: Yes, but if this component is reused somewhere the developer might not be defensive in the higher level component. It is probably better to be defensive in this component because this line will error if an empty array is passed in.

alexkeating (approved):

PR #285, opened 2022-11-04
feature/dao-dashboard-active-weekly-members (PRO-497)

jonathanprozzi:
## Linear Ticket

- [PRO-497](https://linear.app/govrn/issue/PRO-497/[metric]-active-weekly-member-metric)

## Description

- This adds the Weekly Active Members component with the count coming from a 9 week rolling average:
  - Gets contributions from the last 9 weeks with a limit of 500 (we should change this, I think if we move this to the backend we can avoid this -- we need to grab as many results as  meet the criteria so we don't miss a contribution that qualifies)
  - Splits these into 9 week chunks and then grabs each unique person who contributed at least 1 time, with attestations of n=0 for now (we can increase this threshold at some point, but we may want to move the logic onto the backend)
  - Sums the average of unique contributors who meet the criteria and divides by 9 (since 9 week average)
- Added in some additional layout adjustments, but we'll finalize those once all charts are in and we get QA feedback

- Definitely would appreciate some eyes and feedback on this. It may make sense for this to move to the backend

## Screencaptures

### DIFF
--- a/apps/protocol-frontend/src/components/WeeklyActiveMembersShell.tsx
+++ b/apps/protocol-frontend/src/components/WeeklyActiveMembersShell.tsx
@@ -23,12 +23,12 @@ const WeeklyActiveMembersShell = ({ daoId }: WeeklyActiveMembersShellProps) => {
     },
   });
 
-  const setWeeks = () => {
+  const setWeekRange = (weekRange: number) => {
     const weeks = [];
-    for (let i = 0; i <= 8; i++) {
+    for (let i = 0; i <= weekRange - 1; i++) {
       weeks.push(
         contributionsInRangeCount?.filter(
-          (contribution, idx) =>
+          contribution =>
             formatDate(contribution.date_of_engagement) >=
               formatDate(startOfDay(subWeeks(TODAY_DATE, i + 1))) &&
             formatDate(contribution.date_of_engagement) <=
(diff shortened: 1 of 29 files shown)


apps/protocol-frontend/src/components/WeeklyActiveMembersShell.tsx
```diff
@@ -0,0 +1,64 @@
...
+    isLoading,
+    isFetching,
+    isError,
+  } = useContributionList({
+    first: 500, //  TODO: want all contributions in range regardless of count -- better way to do this?
+    where: {
+      guilds: { some: { guild: { is: { id: { equals: daoId } } } } },
+      date_of_engagement: { gte: subWeeks(TODAY_DATE, 9) },
+    },
+  });
```
alexkeating, line 24: I think we should probably handle the aggegation on the backend. It would save fetching all contributions

alexkeating: Prisma upports average aggregation https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing#aggregate. We need to add a method to calculate the average in protocol client using probably this method https://github.com/Govrn-HQ/govrn-monorepo/blob/staging/apps/protocol-api/src/prisma/generated/type-graphql/resolvers/crud/Contribution/AggregateContributionResolver.ts#L10

alexkeating (approved):
We will merge in, but move the logic into the backend as not to block QA

jonathanprozzi:
> We will merge in, but move the logic into the backend as not to block QA

Sounds great! Glad we can merge it in so I can keep refining the layout. Thanks!

PR #291, opened 2022-11-08
Deselect minted contribution

amrro:
## Linear Ticket

[PRO-491](https://linear.app/govrn/issue/PRO-491/minting-minted-contributions)

## Description

Make Select All in Contribution Table deselect the **minted** contributions.


jonathanprozzi:
This works well when I'm testing it -- minted Contributions are being deselected, but I'm noticing that when the infinite scroll loads a new page of Contributions all of the selections are being cleared. This is probably unrelated to this current PR but wanted to note it anyway incase you've seen this

alexkeating:
In this fix it would also be good to fix the issue where all the selected contributions are lost on the next page.

alexkeating (commented):
Select seems to be working but deselect isn't and then the select all is disabled
[Screencast from 11-17-2022 10:25:15 AM.webm](https://user-images.githubusercontent.com/12705146/202487308-39fe0d57-079b-49d5-baed-6130b1bd6f76.webm)

PR #295, opened 2022-11-14
Bug/fix contract sync job

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-523/set-up-contract-job-on-staging-and-prod

## Description

- Make sure contract job is updating properly
- Add pending status
- Log errors in contract sync job
- Update metadata schema

## Considerations
- when launched add pending state
- Have script to identify previous contributions
- And set offset to 0 for first day


libs/protocol-client/src/lib/utils/ipfs.ts
```diff
@@ -4,17 +4,12 @@ import { ContributionV1SchemaCid } from './constants';
...
   name: string;
   details: string;
   proof: string;
   activityName: string;
   image?: string; // keep blank for now
-  govrn: {
-    /* Contribution ID */
-    id: number;
-    activityTypeId: number;
-  };
```
amrro, line 17: May I ask why this is removed?

libs/protocol-client/src/lib/client/contribution.ts
```diff
@@ -164,6 +170,14 @@ export class Contribution extends BaseClient {
   ) {
     const contract = new GovrnContract(networkConfig, signer);
     const transaction = await contract.mint(args);
+    await this.sdk.updateUserOnChainContribution({
+      id: id,
+      status: 'pending',
```
amrro, line 175: I looked for the frontend to check for technically affected code; there are none. 

But I have some observations:
- Shouldn't editing be [disabled](https://github.com/Govrn-HQ/govrn-monorepo/blob/7925c9f35c097fe03c520a6e9803780342464f6a/apps/protocol-frontend/src/components/ContributionsTable.tsx#L205) for `pendig` status too? if editing is enabled, could this duplicate contributions for contract sync job?
- We might need to look for a third [emoji](https://github.com/Govrn-HQ/govrn-monorepo/blob/a724a9c967485dedbac57b247e15655cfc21ab53/apps/protocol-frontend/src/components/ContributionDetailShell.tsx#L71-L73) for `pending`!

apps/contract-sync-job/src/db.ts
```diff
@@ -91,20 +91,62 @@ export const upsertContribution = async (contribution: ContributionData) => {
...
-      },
-    });
+  const existingContribution = await govrn.contribution.list({
+    where: {
+      details: { equals: contribution.details },
+      name: { equals: contribution.name },
+      proof: { equals: contribution.proof },
+      status: { is: { name: { equals: 'pending' } } },
+      user_id: { equals: contribution.user_id },
+      on_chain_id: { equals: null },
```
amrro, line 101: I like this approach for finding out matching contributions under minting 👏🏻.

amrro (approved):
Nice work! Left some observations.

libs/protocol-client/src/lib/client/contribution.ts
alexkeating: Good call, yes we should disable and we can add another emoji

libs/protocol-client/src/lib/utils/ipfs.ts
alexkeating: Yes, unfortunately when testing I realized that id doesn't work when a contribution is being synced across multiple environments and potentially people could target other peoples contributions to be overwritten.

PR #299, opened 2022-11-14
Add text to add dao

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-461/ability-to-request-a-dao-to-be-added

## Screencaptures
Screenshot from 2022-11-14 12-00-10


jonathanprozzi (approved):
🚀

jonathanprozzi (commented):
Oh! One thing I thought of! We should consider also adding this to the EditContributionForm tooltip. Also to the BulkAttribute modal -- basically anywhere that we have a DAO.

alexkeating:
> Oh! One thing I thought of! We should consider also adding this to the EditContributionForm tooltip. Also to the BulkAttribute modal -- basically anywhere that we have a DAO.

Good call outs, I will it there as well. Thought this covered it lol

PR #303, opened 2022-11-15
feature/dao-dashboard-fix-data-cutoff (PRO-589)

jonathanprozzi:
## Linear Ticket

- [PRO-589](https://linear.app/govrn/issue/PRO-589/data-cut-off-on-smaller-screens)

## Description

This adjusts the breakpoint logic for displaying the label layer on the pie charts. The labels are now hidden on laptops with smaller screens. We can revisit this based on feedback but this is the most consistent experience given that sometimes activity type names are long and can get crowded on smaller screens.

Also, seeing an opportunity to refactor the pie charts. Going to do this after the other QA tickets and it'll be in a separate PR.

## Screencaptures

dao-dashboard-pie-charts-removed-labels


apps/protocol-frontend/src/components/ContributionMembersPie.tsx
```diff
@@ -21,7 +21,13 @@ const ContributionMembersPie = ({
...
   isError,
 }: ContributionMembersPieProps) => {
-  const isMobile = useBreakpointValue({ base: true, lg: false });
+  const isSmallerScreen = useBreakpointValue({
+    base: true,
+    md: true,
+    lg: true,
+    xl: false,
+    '2xl': false,
+  });
```
alexkeating, line 30: When testing this seems to make the dashboard view less responsive and cuts off some data

jonathanprozzi: Should be good now -- I used a separate one for the widths than the layers

alexkeating (approved):

PR #307, opened 2022-11-16
use env for Alchemy key

scottrepreneur:
## Linear Ticket

[Use env for Alchemy Key](https://linear.app/govrn/issue/PRO-622/alchemy-key-for-staging)

## Description

Use an environment variable for this key.


scottrepreneur:
Do we need to include a key for the tests?

alexkeating:
> Do we need to include a key for the tests?

I don't think it hurts. Does alchemy support company accounts? The last step is to get the appropriate keys

scottrepreneur:
Yeah definitely

PR #320, opened 2022-11-22
Schema updates for Prisma 4

Flip-Liquid:
Schema updates for Prisma 4 migration

## Linear Ticket

[Pro-332](https://linear.app/govrn/issue/PRO-332/upgrade-to-prisma-4)

## Description

The following changes are being made to the schema:
```
1 Guild ---> MANY TwitterAccount
MANY Tweet ---> MANY Contribution
1 User ---> MANY TwitterUser
```
The following relation is being made explicit: 
```
1 Linear issue ---> 1 Contribution
```


alexkeating (changes requested):
Looks good, one thing I don't see is the prisma 4 update in the package.json and it looks like a bunch of the end to end tests are failing. My guess is we have to generate the gql to resolve that issue.

Flip-Liquid:
@alexkeating these are just the schema changes in this PR. I'll regenerate the client though.

alexkeating:
Looks like prettier step is failing, there should be a command in the package.json to fix this

PR #321, opened 2022-11-22
Fix null usernames

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/PRO-663/show-0x-if-no-user-name

## Description

- Add address for null display name

### DIFF
--- a/apps/protocol-api/src/prisma/resolvers/Contribution.ts
+++ b/apps/protocol-api/src/prisma/resolvers/Contribution.ts
@@ -297,6 +297,9 @@ export class ContributionCountByUser {
     nullable: true,
   })
   display_name?: string;
+
+  @TypeGraphQL.Field(_type => String)
+  address: string;
 }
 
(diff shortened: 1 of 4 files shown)


jonathanprozzi:
Awesome. I'm noticing that it's using this for my name as well even though my name is set. Going to test since I think it's because I may be using `name` -- will check that again.

CleanShot 2022-11-22 at 14 57 30

jonathanprozzi:
Yup, that's exactly what it is -- I have my `name` set in my local db but not my `display_name` -- I think there may be other places where we use name instead of display_name

jonathanprozzi (approved):
Approved -- noting that we may want to think about our use of  `name` and `display_name` since I think we're using `name` at some places in the frontend

alexkeating:
> Approved -- noting that we may want to think about our use of `name` and `display_name` since I think we're using `name` at some places in the frontend

Yea, I think we have a ticket about that

PR #326, opened 2022-11-23
fix/user-name-fallbacks (PRO-667) -- Ready for Re-review and Merge

jonathanprozzi:
## Linear Ticket

- [PRO-667](https://linear.app/govrn/issue/PRO-667/display-nameuser-name-fallbacks)

## Description

- Adds truncated address fallbacks when a user name is null

## Screencaptures

Before:

attestation-no-name
table-no-name
home-shell-no-name
dashboard-no-name
govrn-attestation-dao-name

After:

table-name-fallback
attestation-name-fallback
dashboard-name-fallback
home-shell-name-fallback


apps/protocol-frontend/src/components/HomeShell.tsx
```diff
@@ -86,7 +86,9 @@ const HomeShell = () => {
                   bgGradient="linear(to-l, #7928CA, #FF0080)"
                   bgClip="text"
                 >
-                  {userData?.name || userDataByAddress?.name}
+                  {userData?.name ||
+                    userDataByAddress?.name ||
+                    displayAddress(userData?.address)}
```
alexkeating, line 91: I think we should abstract this or statement into a function. I can imagine us adding more logic such as ENS

jonathanprozzi: Good call. Maybe we can do the fallbacks like this:
User name > ens > truncated address 

Will make a function we can reuse or adjust

alexkeating:
Is this good or are we waiting for the ENS functionality?

jonathanprozzi:
> Is this good or are we waiting for the ENS functionality?

We're waiting. Going to wrap it today

jonathanprozzi:
Ah, wagmi's `useEnsName` hook doesn't support Gnosis out of the box. Will take a look a bit deeper here.

alexkeating:
> Ah, wagmi's `useEnsName` hook doesn't support Gnosis out of the box. Will take a look a bit deeper here.

I think that makes sense, ENS isn't on Gnosis. Maybe, we need to create a mainnet provider to use that hook?

apps/protocol-frontend/src/components/ConnectWallet.tsx
```diff
@@ -112,7 +117,7 @@ const ConnectWallet: React.FC<ConnectWalletProps> = ({ showNetwork }) => {
                       paddingRight={2}
                     >
                       <Text color="gray.800" fontSize="sm">
-                        {name || displayAddress(address || '')}
+                        {userEnsName ?? displayAddress(address)}
```
scottrepreneur, line 120: can you use the hook here too?

scottrepreneur (approved):

jonathanprozzi:
Made the following changes since review and approval:

- https://linear.app/govrn/issue/PRO-777/remove-mainnet-from-wagmi-for-now
- https://linear.app/govrn/issue/PRO-776/use-ens-fetch-outside-wagmi

- Adds in a `useEnsNameEthers` hook that is similar to how wagmi's `useEnsName` works under the hood (with react-query)
- This is utilized in our `useDisplayName` but also used separately in the `ConnectWallet` component since that only uses account or ENS, not the user's display name
- Having this separate allows for us to integrate it into other places like the table where we may want address or ENS only, and not the full display name fallbacks

apps/protocol-frontend/src/components/HomeShell.tsx
```diff
@@ -80,7 +82,10 @@ const HomeShell = () => {
               <Text paddingBottom={8}>
                 Welcome back,{' '}
                 <Text as="span" fontWeight="bolder">
-                  {userData?.name || userDataByAddress?.name}
+                  {displayName}
+                  {/* {userData?.name ||
+                    userDataByAddress?.name ||
+                    displayAddress(userData?.address)} */}
```
alexkeating, line 88: Left over comments

jonathanprozzi: Ah! Thanks 🙏

apps/protocol-frontend/src/utils/web3.ts
```diff
@@ -27,7 +27,7 @@ const gnosisChain: Chain = {
 
 const dev = import.meta.env.MODE || false;
 const defaultChains = dev
-  ? [gnosisChain, chain.goerli, chain.rinkeby, chain.localhost]
+  ? [gnosisChain, chain.mainnet, chain.goerli, chain.rinkeby, chain.localhost]
```
alexkeating, line 30: Don't we need mainnet when dev is false?

jonathanprozzi: Should also remove it from dev if we aren't supporting

jonathanprozzi: I added in the `chain.mainnet` in the prior version of this PR when we used the wagmi hook for ENS. Removing this won't cause confusion in the interim and still lets use ENS

PR #329, opened 2022-11-23
Set user data

alexkeating:
## Linear Ticket

## Description

- Set user data after user is created


jonathanprozzi (approved):
🔥

PR #331, opened 2022-11-29
Average active guild users by parameterized window

Flip-Liquid:
## Linear Ticket

[Add the link to the Linear ticket here.
](https://linear.app/govrn/issue/PRO-587/move-average-logic-to-the-backend)

## Description

Add a custom resolver for calculating the average active users in a guild with a parameterized window size and overall lookback duration.

Calculates the average active members by retrieving the number of distinct contributors to a specific guild and breaking them into a number of buckets of specified width (in days). The number of buckets over which the average should be computed is also specified as an input parameter. 

e.g. a 9w rolling weekly average would be specified as a 7d bucket width looking back over 9 windows worth of data.

## Screencaptures


apps/protocol-api/src/prisma/resolvers/Guild.ts
```diff
@@ -0,0 +1,75 @@
...
+    const windowSizeDays = args.where.windowSizeDays;
+
+    const totalLookbackDays = windowsToLookback * windowSizeDays;
+    const lookbackDate = new Date(startDate.getDate() - totalLookbackDays);
+
+    const result = await prisma.$queryRaw<number>`
+      WITH unique_guild_contributor_daily_count as (
+        SELECT
+          count(distinct c.user_id) as distinct_contributors,
+          EXTRACT(year from gc.createdAt) as year,
```
jonathanprozzi, line 53: I get an error when testing this in graphiQL:
`"message": "\nInvalid `prisma.$queryRaw()` invocation:\n\n\n  Raw query failed. Code: `42703`. Message: `column gc.createdat does not exist`",`

I migrated before testing, but also see that this column does exist in my db. Noticing that the casing of `createdat` is what's coming back in the error but seeing it as `createdAt` in the query itself.

Flip-Liquid: it likely needs to be "quoted" to capture the casing. thank you for catching this!

apps/protocol-api/src/main.ts
```diff
@@ -83,6 +83,7 @@ const permissions = shield(
       linearUsers: hasToken,
       linearIssues: hasToken,
       contributionStatuses: hasToken,
+      getAverageActiveGuildUsers: or(isAuthenticated, hasToken),
```
alexkeating, line 86: Shouldn't this be getGuildUsersAverageByActive

alexkeating, line 86: Shouldn't this be getGuildUsersAverageByActive

apps/protocol-api/src/prisma/resolvers/Guild.ts
```diff
@@ -0,0 +1,75 @@
...
+@TypeGraphQL.InputType('GetActiveUsersInput')
+export class GetActiveUsersInput {
+  @TypeGraphQL.Field(_type => Int)
+  guildId: number;
+
+  @TypeGraphQL.Field(_type => Date)
+  startDate: Date;
+
+  @TypeGraphQL.Field(_type => Int)
+  windowsToLookback: number;
```
alexkeating, line 16: Maybe numWindows would be better?

apps/protocol-api/src/prisma/resolvers/Guild.ts
```diff
@@ -0,0 +1,75 @@
...
+          AND gc."guild_id" = ${guildId}
+        )
+        GROUP BY 
+          year, month, day_bucket
+      )
+
+      SELECT
+        avg(distinct_contributors)
+      FROM
+        unique_guild_contributor_daily_count
```
alexkeating, line 71: Were we unable to use there existing aggregation functionality? Just curious not action item here.

alexkeating (commented):
Some naming updates potentially

jonathanprozzi:
@Flip-Liquid adding the "" around `createdAt` worked!

I'm now able to test this within GraphiQL. However I'm getting back an average of 2.5 for my test DAO which is only me (compared to 0.78 from the original calculation). Will dig a bit deeper and see what's up with that.

Flip-Liquid:
@jonathanprozzi fixed a few more issues. empties and zeroes weren't being handled correctly. ran a few test queries and it seems like it's working properly now

apps/protocol-api/src/main.ts
Flip-Liquid: `By` clauses refer to the grouping of the data. e.g. `..ByDay` queries return the noun bucketed by day.
 
do you feel this naming doesn't conform to what we laid out during the offsite?

Flip-Liquid: @alexkeating

apps/protocol-api/src/prisma/resolvers/Guild.ts
Flip-Liquid: unfortunately not. we're grouping the count of contributors by a derived field, the `day_bucket`. that's not supported in prisma's aggregate/groupBy semantics

apps/protocol-api/src/main.ts
alexkeating: Ah okay that makes sense. Maybe the verb is average instead of get average? As we will have no other verb variants of average. So maybe averageActiveGuildUsers. And what would be averageActiveGuildUsers?

Flip-Liquid: `getContributionCountByActivityType` suggests a pattern that we could follow. maybe `getActiveGuildUsersAverage`? we'd be following e.g. `get...Count|Average|Max`

alexkeating: Yea, I that makes sense verb is get and noun is ActiveGuildUsersAverage.

Flip-Liquid:
@alexkeating update the naming per our conversation

jonathanprozzi (approved):
looks good!

PR #332, opened 2022-11-30
Grabbing some Low lift tickets

twblack88:
## Linear Ticket

 [PRO-675](https://linear.app/govrn/issue/PRO-675/make-the-add-dao-text-bolded-and-new-line)
and 

[PRO-677](https://linear.app/govrn/issue/PRO-677/not-update-text-based-on-time-filter-for-heat-map)

## Description

Making a small attempt on low lift tickets to crack knuckles and get tinkering. Wasn't able to test, so if someone wants to spin up locally and check bold weight that would be rad.  

We might consider:
- [  ] changing the `ReportForm.tsx ` airtable link to a different component. (line 314)
- [  ] revamping date range picker to sit flush with the bar chart, since that's what updates.

### DIFF
--- a/apps/protocol-frontend/src/components/ReportForm.tsx
+++ b/apps/protocol-frontend/src/components/ReportForm.tsx
@@ -308,7 +308,7 @@ const ReportForm = ({ onFinish }: { onFinish: () => void }) => {
             tip={
               <>
                 Please select a DAO to associate this Contribution with. 
-                <Box fontWeight={700}>
+                <Box fontWeight={700} lineHeight={2}>
                   This is optional. Don't see your DAO? Request to add it{' '}
                   <a
                     href="https://airtable.com/shrOedOjQpH9xlg7l"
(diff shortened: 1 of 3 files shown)


jonathanprozzi:
We should make this same update to the `EditContributionForm` since the DAO select field is there as well

apps/protocol-frontend/src/components/ReportForm.tsx
```diff
@@ -306,16 +307,18 @@ const ReportForm = ({ onFinish }: { onFinish: () => void }) => {
...
-                <a
-                  href="https://airtable.com/shrOedOjQpH9xlg7l"
-                  target="_blank"
-                  rel="noopener noreferrer"
-                  style={{ textDecoration: 'underline' }}
-                >
-                  here
-                </a>
+                Please select a DAO to associate this Contribution with. 
+                <Box fontWeight={700}>
```
jonathanprozzi, line 311: We may want to either add a tiny bit of padding or increase the line-height. We can play with this to see what we think

jonathanprozzi: Going to push up a quick change to use Chakra's link and also some prettier fixes

jonathanprozzi (approved):
lgtm @twblack88 🚀

PR #333, opened 2022-11-30
upgrade prisma and prisma client

Flip-Liquid:
## Linear Ticket

[Pro-322](https://linear.app/govrn/issue/PRO-332/upgrade-to-prisma-4)

## Description

Upgrade prisma and prisma client to v 4

## Screencaptures


Flip-Liquid:
@alexkeating src/prisma/resolvers/Contribution.ts L652 is the only area potentially affected by the upgrade. blocked on testing it because of the DB permissions issue

alexkeating:
Okay sounds good, let me know when tested and looks good. I can do a once over and test as well

Flip-Liquid:
@jonathanprozzi mind giving the app a quick spin with this?

apps/protocol-api/src/main.ts
```diff
@@ -89,39 +89,39 @@ const permissions = shield(
...
-      upsertLinearTeam: hasToken,
-      upsertLinearUser: hasToken,
+      upsertOneAttestation: hasToken,
+      upsertOneActivityType: hasToken,
+      upsertOneContribution: hasToken,
+      upsertOneLinearCycle: hasToken,
+      upsertOneLinearIssue: hasToken,
+      upsertOneLinearProject: hasToken,
+      upsertOneLinearTeam: hasToken,
+      upsertOneLinearUser: hasToken,
```
alexkeating, line 124: Damn this breaks our naming convention

package.json
```diff
@@ -44,7 +44,7 @@
     "@nivo/pie": "^0.80.0",
     "@nx-tools/docker-metadata": "^2.2.1",
     "@nx-tools/nx-docker": "^2.3.0",
-    "@prisma/client": "^3.12.0",
+    "@prisma/client": "4",
```
alexkeating, line 47: We should  probably be more specific here and have "^4.7.1"

package.json
```diff
@@ -182,7 +182,7 @@
     "path": "^0.12.7",
     "path-browserify": "^1.0.1",
     "prettier": "2.7.1",
-    "prisma": "^3.12.0",
+    "prisma": "4",
```
alexkeating, line 185: This is a duplicate and can be removed since we have prisma in dependencies

Flip-Liquid: prisma is for the CLI and @prisma/client is the module where the autogenerated client actually sits. I don't think these are duplicates. see https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client

apps/protocol-api/src/prisma/resolvers/Contribution.ts
```diff
@@ -768,6 +788,14 @@ export class ContributionCustomResolver {
         AND gc."guild_id" = ${guildId}
       ) GROUP BY u.display_name, u.id
       ORDER BY count;`;
-    return result;
+    const massagedResult = result.map(x => {
```
Flip-Liquid, line 791: this is bad and i feel bad

jonathanprozzi:
Adding notes from our testing session last night:
- New types from Prisma 4 caused issues with the frontend, particularly the Dashboards and `getContributionCountByDateForUserInRange`
- Type for `count` was changed to a `number` and not a `string` which was the primary type issue
- Queries are working, need to make some slight fixes to the Heat Map charts since they expect the date in a different format -- will address this in a separate PR after this is merged

To do:
- Build is breaking because of type differences for the Dates 
- After merging in, we will address the date issue for the heat maps in an immediate PR (basically, the heat map expects dates in a very particular format (with no time value) and so we have to do a small conversion prior to passing into the heat map. Already tested and it works, but don't want to add it to this PR

@Flip-Liquid I know I omitted a lot of the details from last night -- you crushed the quick fixes!

jonathanprozzi:
That resolved the build errors from the type, but will look into Cypress run

Flip-Liquid:
thanks for the help on the FE changes @jonathanprozzi

jonathanprozzi:
> thanks for the help on the FE changes @jonathanprozzi

For sure! super close, not sure why the Cypress run is failing but I think after we resolve that we can get this merged in!

alexkeating:
> Adding notes from our testing session last night:
> 
>     * New types from Prisma 4 caused issues with the frontend, particularly the Dashboards and `getContributionCountByDateForUserInRange`
> 
>     * Type for `count` was changed to a `number` and not a `string` which was the primary type issue
> 
>     * Queries are working, need to make some slight fixes to the Heat Map charts since they expect the date in a different format -- will address this in a separate PR after this is merged
> 
> 
> To do:
> 
>     * Build is breaking because of type differences for the Dates
> 
>     * After merging in, we will address the date issue for the heat maps in an immediate PR (basically, the heat map expects dates in a very particular format (with no time value) and so we have to do a small conversion prior to passing into the heat map. Already tested and it works, but don't want to add it to this PR
> 
> 
> @Flip-Liquid I know I omitted a lot of the details from last night -- you crushed the quick fixes!
> 
> Update:
> 
>     * I added the heatmap fixes since I pushed something up to resolve the type changes preventing the build. Since I already touched those files any way to resolve the build issue I left the format changes in

+1 Thanks for fixing those before merging

Flip-Liquid:
@jonathanprozzi lets merge before #331, I need to send an iteration on there anyway

alexkeating:
I tried applying the migrations locally and I am getting this error?

Screenshot from 2022-12-08 08-06-15

Flip-Liquid:
That looks like it's when you're running the protocol-api right? Did you
run migrate/generate/generate-gql

alexkeating:
> migrate/generate/generate-gql

Yes, that was with protocol-api. I ran `nx run protocol-api:migrate`. Should

> generate-gql
> [](#)

Ah I think I found the issue. I forgot to run yarn install

Flip-Liquid:
@alexkeating looks like we're good to go here pending your review 💯

alexkeating (approved):
Looks good to me

PR #335, opened 2022-12-01
feature/update-app-landing-page (PRO-604)

jonathanprozzi:
## Linear Ticket

- [PRO-604](https://linear.app/govrn/issue/PRO-604/app-landing-page)
- [PRO-700](https://linear.app/govrn/issue/PRO-700/color-theme-alignments)

## Description

- Adds the new landing page design and makes theme changes
- Holding off on merging until we resolve some open questions on colors, but wanted to see if this passes the Cypress CI run

## Screencaptures

https://user-images.githubusercontent.com/9438776/205138091-723511f9-bacb-4d45-9da1-2eda3e699fe5.mp4

https://user-images.githubusercontent.com/9438776/205138095-1b130faf-f46e-4b27-b914-47b91f5efd25.mp4


amrro:
I think this closes this [ticket](https://linear.app/govrn/issue/PRO-442/remove-underline-decoration-from-social-networks) too.

amrro (commented):
Looks good!

One small thing, the space/lineheight between text is a bit high.
I noticed that two components for the same section are used, adding more space.

jonathanprozzi:
> Looks good!
> 
> One small thing, the space/lineheight between text is a bit high. I noticed that two `<Text>` components for the same section are used, adding more space.

Thanks! That's similar to the spacing in the Figma. I agree though it does seem a bit much. @twblack88 ?

apps/protocol-frontend/src/components/HomeShell.tsx
```diff
@@ -3,13 +3,11 @@ import { Link } from 'react-router-dom';
...
 import { Flex, Button, Text, VisuallyHidden, Heading } from '@chakra-ui/react';
 import ConnectWallet from '../components/ConnectWallet';
-import PageHeading from './PageHeading';
 import CreateUserForm from './CreateUserForm';
-
 import { useAuth } from '../contexts/AuthContext';
 import { useUser } from '../contexts/UserContext';
-import { GOVRN_MOTTO } from '../utils/constants';
 import { GovrnSpinner } from '@govrn/protocol-ui';
+import GovrnTextLogo from './GovrnTextLogo';
```
alexkeating, line 10: (Arbitrary place for comment)

It seems like the user name text needs to be wrapped or shortened

Screenshot from 2022-12-03 00-18-10

alexkeating: The new colors for the dashboard seem really muted and can be tough to read against a white background
Screenshot from 2022-12-03 00-21-04

jonathanprozzi: Ah good catch, I think I know why -- I included the 50 and 100 values for this whereas we previously didn't. The lightest tints may be too washed out

twblack88:
> That's similar to the spacing in the Figma. I agree though it does seem a bit much.

I humbly disagree. There is a lot of space, we'll end up filling it over time. We should try to stick to the design no?

jonathanprozzi:
> 

I think it's more about the spacing between the lines themselves (@amrro mentioned the 2 text stacks)

apps/protocol-frontend/src/components/HomeShell.tsx
jonathanprozzi: Even with removing the 50 and 100 values, it's still really washed. The new palette seems to be much more washed out. I think I can resolve this by using a subset starting at the darker tints, like 300

jonathanprozzi:
Updated to use a subset of the new palette that omits extremes. We can ideally merge this in while we also explore other color solutions for the dashboards. @twblack88

twblack88 (approved):

PR #336, opened 2022-12-05
feature/update-button-styles (PRO-603)

jonathanprozzi:
## Linear Ticket

- [PRO-603](https://linear.app/govrn/issue/PRO-603/button-styling)

## Description

- Adds in several new variants to our theme, most notably the button variants `primary` and `tertiary` -- `secondary` is still in progress and will add once we lock it in
- Added gradients to the theme colors for use in these updates
- Updated the buttons, switches, checkboxes, and toggles to use the new styles

## Screencaptures

Video uploaded to the Linear ticket -- exceeds the upload size here even when compressed


alexkeating:
Couple comments:

1. The styling update doesn't seem to be on the mobile view wallet button
2. The animation on the checkboxs seem really slow and there really isn't too much of a contrast between a disabled one and a checked one besides the check.

jonathanprozzi:
> 

I agree about the checkbox and the contrast -- the contrast is pretty low and there's a decent amount of styling that goes into making the Chakra checkbox look like the design. I'm going to switch back to the native input for now and see if that reduces the lag.

Our Figma was just updated with the new connect button style, so going to add that now

twblack88 (approved):
One more test to see if we can find it but probably simple changes that are easier to catch once it's on staging.

twblack88:
It's [QA's problem now](https://static01.nyt.com/images/2021/04/30/multimedia/30xp-meme/29xp-meme-mediumSquareAt3X-v5.jpg).

jonathanprozzi:
Reverting to ensure we squash and merge -- PR open for reverting and then we can add this back in

PR #345, opened 2022-12-14
fix/local-cypress-tests (PRO-705)

jonathanprozzi:
## Linear Ticket

- [PRO-705](https://linear.app/govrn/issue/PRO-705/cypress-cannot-run-locally)

## Description

- This fixes Cypress tests locally
- Changes the `env` var that detects if we're in local dev mode -- these chains are needed for our testing mock

## Screencaptures

n/a


jonathanprozzi:
This may break the CI run since the VITE_STAGING_MODE is no longer being used -- this uses the dynamic _MODE to detect without being set in `.env` -- will monitor and update if so

PR #349, opened 2022-12-17
Fix attest and mint

alexkeating:
## Linear Ticket
https://linear.app/govrn/issue/PRO-789/mintattest-is-not-working

## Description

- Fix attest and mint before release

### DIFF
--- a/apps/protocol-api/src/prisma/resolvers/Attestation.ts
+++ b/apps/protocol-api/src/prisma/resolvers/Attestation.ts
@@ -177,7 +177,7 @@ export class AttestationResolver {
         id: a.id,
       },
     });
-    if (update.count !== 1) {
+    if (update.count >= 1) {
       throw `Wrong number of rows updated ${update.count} updateUserOnChainAttestation`;
     }
     if (args.status) {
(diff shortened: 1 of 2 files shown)


apps/protocol-api/src/prisma/resolvers/Attestation.ts
```diff
@@ -177,7 +177,7 @@ export class AttestationResolver {
         id: a.id,
       },
     });
-    if (update.count !== 1) {
+    if (update.count >= 1) {
```
jonathanprozzi, line 180: Great catch!

apps/protocol-frontend/src/components/MintModal.tsx
```diff
@@ -151,10 +151,7 @@ const MintModal = ({ contributions, onFinish }: MintModalProps) => {
           variant="primary"
           type="submit"
           onClick={() => {
-            const c = contributions[0];
-            if (contributions && 'original' in c) {
-              mintHandler(contributions);
-            }
+            mintHandler(contributions);
```
jonathanprozzi, line 154: Ah! Great catch on this

jonathanprozzi:
Looks great. Tested locally on Goerli and it works. Tested the following flows:

- Single mint contribution
- Bulk mint 1+ contributions (tested with 2 selected)
- Attest to single contribution

Bulk attest I noticed that the modal closed immediately but it seems that the attestation worked. I'm going to test again and made a ticket to double check the modal closing logic (also noticed it's using the old button -- I missed that one when QA'ing for the button update)

jonathanprozzi (approved):
Great job! See previous comment for details of testing

PR #340, opened 2022-12-09
Add favorite & membership status to UserGuild Table

amrro:
## Linear Ticket

[PRO-687 - Add Favorite and membership status level columns to GuildUser](https://linear.app/govrn/issue/PRO-687/add-favorite-and-membership-status-level-columns-to-guilduser)

## Description
Changes to `GuildUser` Table:
- Add boolean favorite column.
- Add FK column to `GuildMembershipStatus` table, with default value for existing records corresponds to **Supporter** id

PR #352, opened 2022-12-19
Add Guild Import & Integration Types Table

amrro:
## Linear Ticket
[PRO-690 - Add GuildImport table and integration type table](https://linear.app/govrn/issue/PRO-690/add-guildimport-table-and-integration-type-table)


apps/protocol-api/src/prisma/migrations/20221216154936_add_favorite_membership_status/migration.sql
```diff
@@ -3,4 +3,4 @@ ALTER TABLE "GuildUser" ADD COLUMN     "favorite" BOOLEAN NOT NULL DEFAULT false
 ADD COLUMN     "membership_status_id" INTEGER;
 
 -- AddForeignKey
-ALTER TABLE "GuildUser" ADD CONSTRAINT "GuildUser_membership_status_id_fkey" FOREIGN KEY ("membership_status_id") REFERENCES "GuildMembershipStatus"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+ALTER TABLE "GuildUser" ADD CONSTRAINT "GuildUser_membership_status_id_fkey" FOREIGN KEY ("membership_status_id") REFERENCES "GuildMembershipStatus"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
amrro, line 6: This update in order to stop Prisma from removing then re-adding the fk key in a distinct migration.

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -115,7 +116,24 @@ model GuildActivityType {
...
+  updatedAt DateTime @updatedAt @default(now())
+  name String @unique
+  guild_imports GuildImport[]
+}
 
+model GuildImport {
+  id Int @id @default(autoincrement())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt @default(now())
+  authentication_token Int
```
alexkeating, line 131: I think authentication token should be a string

apps/protocol-api/src/prisma/migrations/20221216154936_add_favorite_membership_status/migration.sql
amrro: Yes.

amrro: Actually, I remembered that in the first [migration](https://github.com/Govrn-HQ/govrn-monorepo/pull/340), I manually changed it to `RESTRICT` instead of `SET NULL` to align with some of other migrations.

alexkeating: Can we undo this? the migrations have already been applied on prod and staging. we aren't going to unapply and then reapply them. If we want to make this change we should modify the .prisma file.

alexkeating (approved):

PR #353, opened 2022-12-19
Add create guild user

amrro:
## Linear Ticket

[PRO-684 - Add createGuildUser to protocol client](https://linear.app/govrn/issue/PRO-684/add-createguilduser-to-protocol-client)


libs/protocol-client/src/lib/client/user.ts
```diff
@@ -38,8 +39,11 @@ export class User extends BaseClient {
 }
 
 class GuildUser extends BaseClient {
+  public async create(args: CreateGuildUserMutationVariables) {
+    return await this.sdk.createGuildUser(args);
```
alexkeating, line 43: Don't we need to add this to the graphql-shield? And also we probably want to have some permissioning on this so that only a user can add themselves to a guild. We also need to prevent the user from adding a membership status. We should take the default status.

amrro: Just noticed that it merged to prod. Thanks for reminding me!

amrro:
I tried to do it with `ownsData` rule, But no clear basis or definition to based on. 
Here er are using _user id_, in other part of the project we are using _address_

alexkeating:
> I tried to do it with `ownsData` rule, But no clear basis or definition to based on. Here er are using _user id_, in other part of the project we are using _address_

OwnsData is deprecated and shouldn't be used. It is the same as allow

apps/protocol-api/src/main.ts
```diff
@@ -105,6 +105,7 @@ const permissions = shield(
       createUserAttestation: and(ownsData, isAuthenticated),
       createUserContribution: and(ownsData, isAuthenticated),
       createUserCustom: or(hasToken, and(ownsData, isAuthenticated)),
+      createGuildUserCustom: or(hasToken, and(ownsData, isAuthenticated)),
```
alexkeating, line 108: Can we remove both has token and ownsData here? A job isn't calling this and ownsData is essentially allow

apps/protocol-api/src/prisma/resolvers/UserGuild.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const user = await prisma.user.findFirst({
+      where: { address: { equals: address } },
+    });
+
+    if (user?.id !== args.data.userId) {
+      throw new Error('Signature address does not equal requested address');
+    }
+
+    return await prisma.guildUser.create({
```
alexkeating, line 43: When creating a user let's give them recruit status instead of null for now

alexkeating (approved):

PR #354, opened 2022-12-19
add discord notification schema

Flip-Liquid:
## Linear Ticket

https://linear.app/govrn/issue/PRO-581/botnotification-setting-schema-design

## Description

Add schema fields for supporting notifications sent to discord users.

## Screencaptures


apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -50,6 +50,26 @@ model DiscordUser {
   discord_id String @unique
   user_id Int
   user User @relation(fields: [user_id], references: [id])
+
+  // notifications
+  // enable notifications
+  enable_notifications Boolean @default(false)
+
+  // submit contributions
+  enable_submit_reminder_period Boolean @default(false)
```
alexkeating, line 59: These look good, on the the enable columns like this it may be clearer to remove the _period suffix. So instead of the above it should be enable_submit_reminder

PR #356, opened 2022-12-21
Add update guild

amrro:
## Linear Ticket

[PRO-686 - Add updateGuild to protocol client](https://linear.app/govrn/issue/PRO-686/add-updateguild-to-protocol-client)

## Description

- Adding a custom mutation `updateGuild` to the protocol client. Only the following fields can be updated: `name`, `discord_id`, `congrats_channel`, `contribution_reporting_channel` and `logo`. 
- fields' names matches the one in the db in order to work.
- Only admins can use this.

### Usage
```ts
try {
  await govrn.guild.update(
    {
      name: 'GOVRNNN',
    },
    { guildId: 13 },
  );
} catch (e) {
  console.dir(e);
}
```


apps/protocol-api/sql/add_membership_status.sql
```diff
@@ -1,5 +1,6 @@
 INSERT INTO "GuildMembershipStatus" (name)
-VALUES ('Member'),
+VALUES ('Admin'),
```
amrro, line 2: Based on this [doc](https://hackmd.io/SDOyv9B4RCaNiLa1MpWuaQ), I added the Admin role.
If this is not the case, I can undo this commit 606b83f978048624d2f3948446fb283cac4aeedd.

apps/protocol-api/src/prisma/resolvers/Guild.ts
```diff
@@ -57,15 +131,15 @@ export class GuildCustomResolver {
...
           FROM generate_series(${lookbackDate}, ${startDate}, '1 day'::interval) dt
-        ) d 
+        ) d
         LEFT JOIN "Contribution" as c
           ON c.date_of_engagement::date = d.dt::date
-        LEFT JOIN "GuildContribution" as gc 
+        LEFT JOIN "GuildContribution" as gc
           ON gc."contribution_id" = c.id
         WHERE (
           gc."guild_id" IS NULL OR gc."guild_id" = ${guildId}
```
amrro, line 140: **Side note**: 
this field is declared as non-null in prisma's schema. Is this check necessary?

alexkeating: Yes, because a contribution might not have a guild_id

alexkeating (approved):

PR #358, opened 2023-01-04
Feature/csv uploader

alexkeating:


apps/jobs/member-csv-upload-job/src/main.ts
```diff
@@ -0,0 +1,117 @@
...
+  // pull
+  // transform
+  // enqueue
+  // etc
+  const pullTransform = async (conn: NatsConnection, msg: JsMsg) => {
+    const sc = StringCodec();
+    const data = sc.decode(msg.data);
+    console.log('processing message...');
+    // DAO ID, address, discord name/username (optional), dicord_id (optional	), admin
+    const [daoId, address, discordName, discordId, admin] = data.split(',');
```
alexkeating, line 38: This was probably wrong in the ticket but we should probably remove id

alexkeating (commented):
But looks fine so far, we can probably follow on refactor to reorganize the code

PR #359, opened 2023-01-04
feature/custom-toast-variants (PRO-711)

jonathanprozzi:
## Linear Ticket

- https://linear.app/govrn/issue/PRO-711/toastsalerts-variants
- https://linear.app/govrn/issue/PRO-767/success-variant
- https://linear.app/govrn/issue/PRO-768/error-variant
- https://linear.app/govrn/issue/PRO-795/politely-request-every-capital-contribution-to-lowercase-instead
  - Many of these updates were in the Toasts, so also including this ticket in the PR 

## Description

This updates our Toast to use a custom component based on the Figma designs (roughly -- had to make a few changes) and moves it to the UI library for reusability across other apps/projects. The use of our Toast component is largely the same as before, but we now can optionally pass in a title or not. I left the titles in since our current alert messages rely on them, but we can now remove if we like.

Since it's now in the design system, you'll want to import `{ useGovrnToast }` from there.

This adds structure for us to include custom Toasts for additional uses (info, loading, warning) but only includes the new styles for the success and error as that's what we have so far in the Figma design update.

Additionally, we can optionally close the Toasts -- when creating your Toast you can add the `isClosable: false` to the Toast config when calling it (include in the object passed into `toast.success` for example)

## Screencaptures

error-toast-no-title
error-toast-title
success-toast-no-title
success-toast-title


apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -173,13 +173,13 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
       });
       await getUser();
       toast.success({
-        title: 'User Profile Updated',
+        // title: 'User Profile Updated',
```
alexkeating, line 176: Remove rather than comment these

apps/protocol-frontend/src/hooks/useAttestationBulkMint.ts
```diff
@@ -49,7 +49,7 @@ const useAttestationBulkMint = () => {
 
         if (results.length > 0) {
           toast.success({
-            title: 'Attestation successfully minted',
+            title: 'Attestation Successfully Minted',
```
alexkeating, line 52: Does successfully and minted need to be capitalized

apps/protocol-frontend/src/contexts/UserContext.tsx
jonathanprozzi: Oops! Meant to remove that entirely after I added the title back in.

apps/protocol-frontend/src/hooks/useAttestationBulkMint.ts
jonathanprozzi: Yup, the title should be capitalized. Will fix

apps/protocol-frontend/src/hooks/useLogout.ts
```diff
@@ -18,7 +19,7 @@ const useLogout = () => {
     } catch (err) {
       console.error(err);
       toast.error({
-        title: 'Failed to logout user',
+        // title: 'Failed to logout user',
```
alexkeating, line 22: One more straggler

PR #367, opened 2023-01-16
feature/dao-membership-profile-page (PRO-616)

jonathanprozzi:
## Linear Ticket

- [PRO-616](https://linear.app/govrn/issue/PRO-616/dao-card-on-profile-page)
- [PRO-779](https://linear.app/govrn/issue/PRO-779/states)
- [PRO-617](https://linear.app/govrn/issue/PRO-617/show-daos-youve-joined)
- [PRO-817](https://linear.app/govrn/issue/PRO-817/create-feature-flag-component)

## Description

- Adds in the UI for the _My DAOs_ section of the Profile
- Modifies our Select components to support better portalling and respect to clickable elements behind the open menu
- Adds the `ProfileDaos` and `DaoCard` components
- Fetching is done at the `ProfileDaos` component so that we don't block any loading for the rest of the Profile (similar to how we did this for the DAO Dashboards)
- Uses mock data that'll be replaced:
  - Noted throughout the code but the mock data is for the user's DAOs that (mock data being used for the `favorite` and their `role`)  

## Screencaptures

govrn-profile-daos-mobile
govrn-profile-daos

### DIFF
--- a/apps/protocol-frontend/src/components/DaoCard.tsx
+++ b/apps/protocol-frontend/src/components/DaoCard.tsx
@@ -1,0 +1,62 @@
+import { Flex, Heading, Text } from '@chakra-ui/react';
+
+type DaoRoles = 'admin' | 'contributor' | 'recruit';
+
+type Dao = {
+  id: number;
+  name: string;
+  role: DaoRoles;
+  isFavorited: boolean;
+};
+
+interface DaoCardProps {
+  dao: Dao;
+}
+
+const daoBgGradient = {
+  admin: 'linear-gradient(180deg, #5100E4 0%, #9766EF 100%)',
+  contributor: 'linear-gradient(180deg, #DCCCFA 0%, #F8F4FF 100%)',
+  recruit: 'linear-gradient(180deg, #E2E8F0 0%, #F7FAFC 100%)',
+};
+
+const daoNameColor = {
+  admin: 'white',
+  contributor: 'brand.purple',
+  recruit: 'gray.700',
+};
+
+const DaoCard = ({ dao }: DaoCardProps) => {
+  return (
+    <Flex
+      direction="column"
+      alignItems="center"
+      justifyContent="center"
+      bg={daoBgGradient[dao.role]}
+      borderWidth="1px"
+      borderStyle="solid"
+      borderColor={dao.role === 'recruit' ? 'gray.200' : 'brand.purple'}
+      minWidth="14rem"
+    >
+      <Flex
+        direction="column"
+        alignItems="center"
+        justifyContent="center"
+        minHeight="6rem"
+      >
+        <Heading
+          as="h4"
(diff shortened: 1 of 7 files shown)


alexkeating:
One UI thing I noticed, maybe a gap should be enforced
Screenshot from 2023-01-16 14-53-59

apps/protocol-frontend/src/components/DaoCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+            ) : (
+              <IconButton
+                aria-label="Click on the star to favorite this DAO."
+                bg="transparent"
+                _hover={{ bg: 'transparent' }}
+                _active={{ bg: 'transparent' }}
+                icon={
+                  dao.favorite === true ? (
+                    <HiStar
+                      fill={dao.role === 'recruit' ? 'gray.700' : '#5100E4'}
```
alexkeating, line 120: This turnary should probably be a constant to be dryer

apps/protocol-frontend/src/components/ProfileDaos.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+// only here for mocking purposes, will replace
+type DaoRoles = 'admin' | 'contributor' | 'recruit';
+
+type Dao = {
+  id: number;
+  name: string;
+  role: DaoRoles;
+  favorite: boolean;
+};
```
alexkeating, line 16: Some code duplication here, with dao card

alexkeating (commented):
Looks good maybe some small nits

apps/protocol-frontend/src/components/DaoCard.tsx
jonathanprozzi: Good calls on all of these especially the gap -- will resolve.

jonathanprozzi:
With gap more explicitly set:

dao-grid-gap

apps/protocol-frontend/src/components/ProfileDaos.tsx
jonathanprozzi: Oh yeah -- figure once we remove the mockData that'll use the same type in both places (likely our UIGuild type)

amsos:
@jonathanprozzi looking good

On the dropdown list to the left of the `join` button, I click the carrot, it shows me daos I can join, then I click the DAO, click the join button, and then it adds me as a `recruit`.

Is that the right flow?

jonathanprozzi:
> @jonathanprozzi looking good
> 
> 
> 
> On the dropdown list to the left of the `join` button, I click the carrot, it shows me daos I can join, then I click the DAO, click the join button, and then it adds me as a `recruit`.
> 
> 
> 
> Is that the right flow?

Thanks! That is the flow. We have a separate ticket for that integration but that is how it'll work.

PR #108, opened 2022-07-23
Feautre/show linear

alexkeating:
## Linear Ticket
https://linear.app/govrn/issue/PRO-264/setup-govrn-linear-integration-for-live-app

## Description

- Switch linear integration to require oauth
- Added to fields to linear user to store access token
- Added typing for prisma client in the server which resolved a bug in one of our queries
- Created AuthContext to hold auth state and logic
- 

## Screencaptures
Screenshot from 2022-07-23 10-57-36
Screenshot from 2022-07-23 10-58-03

PR #368, opened 2023-01-16
Set up Discord oauth endpoint

amrro:
## Linear Ticket

[ENG-730 - Set up Discord oauth endpoint](https://linear.app/govrn/issue/ENG-730/set-up-discord-oauth-endpoint)

## Description
- Related #108 
- Added new columns to `DiscordUser`: `access_token` & `active_token`.
- Two new APIs added: 
   - `\discord_nonce`.
   - `\discord\oauth`.
- Fix not able to disconnect linear; having `linear_users` on user data doesn't mean user is connected. Instead, checking `active_token` flag is added, same applied to `discord_users`.
- Fix `useDisplayName` taking too long in order to load because of dependency on ENS name fallback.
- Added Icons to linear & discord buttons.

### Required .env variables
```
DISCORD_REDIRECT_URI=<>
DISCORD_CLIENT_ID=<>
DISCORD_CLIENT_SECRET=<>
```

### Ready to use example
```ts
  const handleDiscordAuth = async () => {
    const res = await fetch(`${BACKEND_ADDR}/discord_nonce`, {
      method: 'GET',
      credentials: 'include',
    });
    const state = await res.text();

    const params = new URLSearchParams();
    params.append('client_id', DISCORD_CLIENT_ID);
    params.append('redirect_uri', REDIRECT_URL);
    params.append('response_type', 'code');
    params.append('scope', 'identify guilds.join');
    params.append('state', `${state}/${userData?.address}`); // generate string to prevent crsf attack
    params.append('prompt', 'consent');
    window.location.href = `https://discord.com/oauth2/authorize?${params.toString()}`;
  };

```


apps/protocol-api/src/main.ts
```diff
@@ -535,5 +542,66 @@ app.get('/linear/oauth', async function (req, res) {
   }
 });
 
+app.get('/discord_nonce', async function (req, res) {
+  const nonce = generateNonce();
+  req.session.linearNonce = nonce;
+  res.setHeader('Content-Type', 'text/plain');
+  res.status(200).send(req.session.nonce);
+});
```
alexkeating, line 550: Why do we need a separate nonce endpoint? Why doesn't the other nonce suffice?

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -13,14 +13,20 @@ import useDisplayName from '../hooks/useDisplayName';
...
 import ProfileDaos from './ProfileDaos';
+import { CgLinear, SiDiscord } from 'react-icons/all';
+import useDiscordUserDisconnect from '../hooks/useDiscordUserDisconnect';
 
 const LINEAR_CLIENT_ID = import.meta.env.VITE_LINEAR_CLIENT_ID;
 const LINEAR_REDIRECT_URI = import.meta.env.VITE_LINEAR_REDIRECT_URI;
 const BACKEND_ADDR = `${BASE_URL}`;
 
+const DISCORD_CLIENT_ID = '1062060045929549844';
+const REDIRECT_URL = 'http://localhost:4000/discord/oauth';
```
alexkeating, line 24: These should be environment variables so we can use this in multiple environments.

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -74,6 +80,34 @@ const ProfileForm = () => {
     });
   }, [userData, disconnectLinearUser]);
 
+  const disconnectDiscordOnClick = useCallback(async () => {
```
alexkeating, line 83: We should wrap this logic in the hook? I don't think we need the useCallback here

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -74,6 +80,34 @@ const ProfileForm = () => {
...
+      return;
+    }
+    await disconnectDiscordUser({
+      userId: userData?.id,
+      username: userData?.name || '',
+      discordUserId: userData?.discord_users[0].id,
+    });
+  }, [userData, disconnectDiscordUser]);
+
+  const handleDiscordAuth = async () => {
```
alexkeating, line 94: Same here this should be wrapped in a hook

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -165,11 +200,43 @@ const ProfileForm = () => {
...
+              align="flex-start"
+              marginTop={4}
+              marginBottom={4}
+              gap="20px"
+              width={{ base: '100%', lg: '50%' }}
+            >
+              <Heading as="h4" size="sm" fontWeight="medium" color="gray.700">
+                Discord
+              </Heading>
+              {userData?.discord_users?.length &&
```
alexkeating, line 220: Can we asbstract this outside of the JSX. Things are a bit complex here

apps/protocol-frontend/src/hooks/useDisplayName.ts
```diff
@@ -9,18 +9,17 @@ const useDisplayName = () => {
...
-    userEnsName ||
-    displayAddress(address);
-  return { displayName };
+    userDataByAddress?.display_name;
+
+  if (isLoading) {
+    return { displayName };
+  }
+
+  return { displayName: displayName || userEnsName || displayAddress(address) };
```
alexkeating, line 22: Why did this need to be changed?

jonathanprozzi:
Awesome! Going through and testing now. 
When I go through the flow it's giving me a 500 error on the redirect so I likely need to set some additional env vars

alexkeating (commented):
When testing I got a 500, and the server side error was `Invalid prisma.discordUser.upsert() invocations`

apps/protocol-frontend/src/hooks/useDisplayName.ts
```diff
@@ -9,18 +9,17 @@ const useDisplayName = () => {
...
-
   const displayName =
     userData?.name ||
     userData?.display_name ||
     userDataByAddress?.name ||
-    userDataByAddress?.display_name ||
-    userEnsName ||
-    displayAddress(address);
-  return { displayName };
+    userDataByAddress?.display_name;
```
jonathanprozzi, line 16: Not sure if it's only me, but this is delaying the render of the entire connect wallet component while it resolves (comparing to what we currently have on prod)

I'll continue to test this

apps/protocol-frontend/src/hooks/useDiscordUserDisconnect.ts
```diff
@@ -0,0 +1,47 @@
...
+        queryClient.invalidateQueries(['userGet']);
+        queryClient.invalidateQueries(['userByAddressGet']);
+        toast.success({
+          title: 'Disconnected discord user',
+          description: 'Your discord user has been disconnected',
+        });
+      },
+      onError: error => {
+        toast.error({
+          title: 'Failed to disconnect discord user',
```
jonathanprozzi, line 38: minor nitpick: we may want to change to "Failed to disconnect Discord user."

apps/protocol-frontend/src/hooks/useDiscordUserDisconnect.ts
```diff
@@ -0,0 +1,47 @@
...
+      });
+
+      return resp;
+    },
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries(['userGet']);
+        queryClient.invalidateQueries(['userByAddressGet']);
+        toast.success({
+          title: 'Disconnected discord user',
```
jonathanprozzi, line 32: Same here -- "Disconnected Discord user."

apps/protocol-frontend/src/components/ProfileForm.tsx
jonathanprozzi: Great call -- can then reuse these env vars in the Signature flow standalone page 🔥

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -152,7 +186,8 @@ const ProfileForm = () => {
               <Heading as="h4" size="sm" fontWeight="medium" color="gray.700">
                 Linear
               </Heading>
-              {userData?.linear_users && userData.linear_users.length > 0 ? (
+              {userData?.linear_users?.length &&
+              userData.linear_users[0].active_token ? (
```
jonathanprozzi, line 190: Ah! So when a user disconnects do we then remove the `active_token`?

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -165,11 +200,43 @@ const ProfileForm = () => {
                   variant="primary"
                   type="submit"
                   onClick={handleLinearOauth}
+                  leftIcon={<CgLinear />}
```
jonathanprozzi, line 203: Nice improvement!

apps/protocol-frontend/src/components/ProfileForm.tsx
amrro: `active_token` is set to false upon disconnecting;

apps/protocol-api/src/main.ts
amrro: we have `nonce` & `linear_nonce`, I thought that we should have nonce for every usage.

alexkeating: Ah okay I see, yea it makes sense to have another

apps/protocol-api/src/sessions.d.ts
```diff
@@ -7,5 +7,6 @@ declare module 'express-session' {
     siwe: SiweResponse;
     nonce?: string;
     linearNonce?: string;
+    discordNonce?: string;
```
amrro, line 10: @alexkeating, what is the purpose of these session variables?

jonathanprozzi:
Awesome! Tested again and the flow works and is very smooth (both connect and disconnect)

jonathanprozzi:
Noticing one small thing when testing again for integrating into the signature flow --

When I click 'Disconnect Discord' and I'm redirected back to the Profile / Profile refreshes, there is a ? added to the URL:

disconnect-discord-url

jonathanprozzi:
> **Update: The Linear disconnect does the same, so it's not specifically related to this PR.**
> 
> Noticing one small thing when testing again for integrating into the signature flow --
> 
> When I click 'Disconnect Discord' and I'm redirected back to the Profile / Profile refreshes, there is a ? added to the URL:
> 
> disconnect-discord-url [ disconnect-discord-url ](https://user-images.githubusercontent.com/9438776/218149632-1f69165b-5be3-4a78-a81c-86ddec06f6df.gif) [ ](https://user-images.githubusercontent.com/9438776/218149632-1f69165b-5be3-4a78-a81c-86ddec06f6df.gif)

In testing the disconnect a bit more, I'm not sure it's going completely through the intended flow. The disconnect calls the `useDiscordDisconnect` hook which is a mutation with an `onSuccess` and `onError` toast -- I'm not seeing the Toast. Instead I'm redirected to the page (or it refreshes) and the URL changes, which seems to be from the `res.status(200).redirect(PROTOCOL_FRONTEND + '/#/profile');` in the Discord oAuth express route. Not sure if that's intended since it looks like we're using the hook to disconnect. This may be my misunderstanding though.

amrro:
> **Update: The Linear disconnect does the same, so it's not specifically related to this PR.**
> 
> Noticing one small thing when testing again for integrating into the signature flow --
> 
> When I click 'Disconnect Discord' and I'm redirected back to the Profile / Profile refreshes, there is a ? added to the URL:


I'm noticing the url changed from `/#/profile` to `/?#/profile`, as if somewhere a search params is being added or some routing issue.

apps/protocol-api/src/sessions.d.ts
alexkeating: This nonce should be passed into the state to prevent crsf attacks

apps/protocol-frontend/src/components/ProfileForm.tsx
```diff
@@ -191,7 +169,7 @@ const ProfileForm = () => {
                 <Button
                   variant="secondary"
                   type="submit"
-                  onClick={disconnectLinearOnClick}
+                  onClick={async () => await disconnectLinearUser()}
```
alexkeating, line 178: You can call this as disconnectLinearUser. No need to create an async function. Same for the function below

amrro: They are not assignable to onClick.

PR #371, opened 2023-01-17
Upgrade to yarn 3

alexkeating:
## Description

- Upgrades us to the latest yarn version


jonathanprozzi (approved):
Looks good. We should also make sure our local node versions match too after this is merged in

PR #372, opened 2023-01-17
DAO Text Area Input (ENG-611)

jonathanprozzi:
## Linear Ticket

- [ENG-611](https://linear.app/govrn/issue/ENG-611/textarea-input-form)

## Description

- This is the UI component for adding DAOs manually via a comma-separated list of addresses
- It'll be accessed on the upcoming DAO Settings page, but for now the modal is temporarily on the Profile page
  - Can be triggered for local testing by using the React DevTools Component tool, selecting the Overlay Context, and setting the `daoMemberAddresses` to `true` -- I did this instead of adding a button to trigger it that we'd need to remove
- Validation checks for valid addresses, that there is at least 1 added, and then removes the duplicates
- Triggers onBlur and onSubmit -- submit handler will be added in the integration ticket 

- There are likely some validation improvements we can do to improve the onBlur (such as if a user clears the input, which you can briefly see in the example) but I also want to open that up for QA and iteration when we get these all in holistically
- 
## Screencaptures

https://user-images.githubusercontent.com/9438776/212960826-7b35f6db-cbc0-4200-8f63-0f0ff9a82bdc.mp4


jonathanprozzi:
Going to update with a basic page view at the intended route so we can anchor this modal.

jonathanprozzi:
create-dao-wip

Added this page that we'll need to include any way. This can be accessed directly via `/dao/create` with the feature prefix. The link will be added to the nav sidebar in a future PR since we're still determining a few UX details with it.

There's improvements we can do with the validation UX, but I think we can continue to improve that when wiring up the import functionality. On pressing the create button, this parses the addresses to be unique values that we can then use to create the DAO

apps/protocol-frontend/src/components/DaoTextareaForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+    e: React.ChangeEvent<HTMLTextAreaElement>,
+  ) => {
+    if (validationError !== null) {
+      setValidationError(null);
+    }
+    setDaoMemberAddresses(e.target.value);
+    validateDaoMemberAddresses();
+  };
+
+  const validateDaoMemberAddresses = () => {
```
alexkeating, line 23: This fails if there is a single address and no comma. It says invalid address. maybe the message should be different.

The same error occurs if the last address does not end in a comma

jonathanprozzi: Ah yeah. Going to rework the validation - I'm not happy with it and it's finicky as it is. I have a version I'm trying that does use RHF and a custom yup validator instead since this ended up being likely  more work to not use it

jonathanprozzi: Updated to use RHF and the validation is smoother. I was going to do that originally but thought we'd run into issues with the validation. The custom yup `.test` is working to validate each address in the list

PR #373, opened 2023-01-17
API Endpoint for import History

amrro:
## Linear Ticket
[ENG-665 - API Endpoint for import History](https://linear.app/govrn/issue/ENG-665/api-endpoint-for-import-history)

## Description
- Add `ImportStatus` enum to `GuildImport` table.
- Link `GuildUser` to the `GuildImport` table.
- Rename `membershipStatus` -> `membership_status` to align with naming
  convention & update usages.
- Add import history API.

### Usage
```ts
    govrnProtocol.guild.import
      .list({})
      .then(value => {
        console.log(value);
      })
      .catch(e => console.error(e));
```


apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -106,8 +112,10 @@ model GuildUser {
   guild_id Int
   guild Guild @relation(fields: [guild_id], references: [id])
   membership_status_id Int?
-  membershipStatus GuildMembershipStatus? @relation(fields: [membership_status_id], references: [id])
+  membership_status GuildMembershipStatus? @relation(fields: [membership_status_id], references: [id])
```
alexkeating, line 115: This will rename the column and we may want to avoid that for now

libs/protocol-client/src/lib/client/guild.ts
```diff
@@ -2,15 +2,24 @@ import { BaseClient } from './base';
...
 } from '../protocol-types';
 import {
   CreateGuildMutationVariables,
   GuildWhereUniqueInput,
 } from '../protocol-types';
+import { GraphQLClient } from 'graphql-request';
 
 // TODO: Add users query
 export class Guild extends BaseClient {
+  imports: GuildImport;
```
alexkeating, line 16: This isn't an array and should be singular

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -19,6 +19,12 @@ enum GuildStatus {
   ONBOARDED
 }
 
+enum ImportStatus {
+  Pending
+  Completed
+  Failed
+}
```
alexkeating, line 26: Oh let's prefer a separate table rather than an enum. That is what we have done for contributions and attestations and it is important to be consistent

alexkeating (commented):
One last request :pray:

PR #376, opened 2023-01-18
Add list `GuildUser` API

amrro:
## Linear Ticket

[ENG-826 - Verify or add guild user endpoint](https://linear.app/govrn/issue/ENG-826/verify-or-add-guild-user-endpoint)


libs/protocol-client/src/lib/graphql/queries.graphql
```diff
@@ -212,9 +212,19 @@ fragment GuildUserFragment on GuildUser {
...
+    id
+    createdAt
+    updatedAt
+    discord_id
+    name
+    congrats_channel
+    logo
+    contribution_reporting_channel
+    status
+  }
```
alexkeating, line 227: I don't think we need all of these fields maybe just id, discord_id, name, logo, and status

apps/protocol-api/src/main.ts
```diff
@@ -74,6 +74,7 @@ const permissions = shield(
       getUser: isAuthenticated,
       guild: or(isAuthenticated, hasToken),
       guilds: or(isAuthenticated, hasToken),
+      guildUsers: or(isAuthenticated, hasToken),
```
alexkeating, line 77: Only users in the guild should be able to fetch other users in the guild. This will require a custom endpoint.

alexkeating: We can handle this in another ticket

alexkeating (approved):

PR #380, opened 2023-01-24
Add attribution

alexkeating:
#### Description
- Click into attribution


jonathanprozzi:
Works great!

Only comment is if we want to add the same ability to view these from the _My Attestations_ tab. The table for that is `MyAttestationsTable`

PR #387, opened 2023-01-26
feature/join-dao-functionality (ENG-863)

jonathanprozzi:
## Linear Ticket

- [ENG-863](https://linear.app/govrn/issue/ENG-863/join-dao-mutation-integration-with-ui)

## Description

- Adds in a `useDaoUserCreate` hook
- Adds user to the selected DAO as a `recruit` and invalidates all related queries

## Screencaptures

https://user-images.githubusercontent.com/9438776/214767493-db47077f-e2b7-4df9-b28a-c0b60ad63fe4.mp4

clears select on join:

govrn-dao-join-clear-select


apps/protocol-frontend/src/hooks/useDaoUserCreate.ts
```diff
@@ -0,0 +1,44 @@
...
+      return mutationData;
+    },
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries(['activityTypes']); // invalidate the activity types query -- covers all args
+        queryClient.invalidateQueries(['userDaos']); // invalidate the userDaos query -- covers all args
+        queryClient.invalidateQueries(['contributionList']);
+        queryClient.invalidateQueries(['contributionInfiniteList']);
+        queryClient.invalidateQueries(['ContributionGetCountYear']);
+        queryClient.invalidateQueries(['contributionGetCount']);
```
alexkeating, line 28: Why are we invalidating all these contribution queries? Shouldn't this really be userDaos only?

jonathanprozzi: I was going back and forth about this. The reason I landed on invalidating these is because they'd be impacted by having a user have a new DAO added to their list. I wasn't completely confident it would be invalidated if we only did the user's DAOs.

PR #388, opened 2023-01-26
Remove contribution form and backend

alexkeating:
## Description

- Remove contribution form and backend

https://linear.app/govrn/issue/ENG-214/deprecate-old-reporting-form


jonathanprozzi (approved):
Looks good! Tested all our existing flows to be safe. 

Feels big that we're removing this! 🚀

PR #391, opened 2023-01-27
feature/dao-manual-create-functionality (ENG-843)

jonathanprozzi:
## Linear Ticket

- [ENG-843](https://linear.app/govrn/issue/ENG-843/manual-dao-create-functionality)
- [ENG-866](https://linear.app/govrn/issue/ENG-866/add-name-to-the-create-a-dao-manual-text-input-modal)

## Description

- Updates our guild user create resolver to allow for creating DAOs in addition to setting status
- When a user joins the DAO via the Create Manual DAO and enters comma separated list of addresses the DAO is created and then each person is added as a recruit. The Summoner is set as an admin (these will be updatable in the Settings page)

## Screencaptures

https://user-images.githubusercontent.com/9438776/215204315-d50b9cef-b129-4f6c-88ab-cc8535a14139.mp4

### DIFF
--- a/apps/protocol-api/src/prisma/resolvers/UserGuild.ts
+++ b/apps/protocol-api/src/prisma/resolvers/UserGuild.ts
@@ -67,6 +67,8 @@ export class GuildUserCustomResolver {
   ) {
     const address = req.session.siwe.data.address;
 
+    console.log('args', args);
+
     const user = await prisma.user.findFirst({
       where: { address: { equals: address } },
     });
--- a/apps/protocol-api/src/prisma/resolvers/UserGuild.ts
+++ b/apps/protocol-api/src/prisma/resolvers/UserGuild.ts
@@ -114,10 +114,6 @@ export class GuildUserCustomResolver {
       throw new Error('Signature address does not equal requested address');
     }
 
-    // modify to fetch the guild and if the guild doesnt exist then skip this check
-    // create the guild in separate query and use this guild id
-    // name would be required in the create guild query
-
     if (args.data.guildId !== undefined) {
       return await prisma.guildUser.create({
         data: {
--- a/apps/protocol-api/src/prisma/resolvers/UserGuild.ts
+++ b/apps/protocol-api/src/prisma/resolvers/UserGuild.ts
@@ -71,6 +71,14 @@ export class GuildUserCustomResolver {
       where: { address: { equals: address } },
     });
 
+    console.log('address', address);
+    console.log('args.data.userId', args.data.userId);
+    console.log('guild id', args.data.guildId);
+
+    if (args.data.guildId === undefined && user?.id !== args.data.userId) {
+      throw new Error('Signature address does not equal requested address');
+    }
+
     if (args.data.guildId === undefined) {
       const guildCreate = await prisma.guild.create({
         data: {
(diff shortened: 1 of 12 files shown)


apps/protocol-api/src/prisma/resolvers/UserGuild.ts
```diff
@@ -54,19 +67,69 @@ export class GuildUserCustomResolver {
...
     const address = req.session.siwe.data.address;
 
+    console.log('args', args);
+
     const user = await prisma.user.findFirst({
       where: { address: { equals: address } },
     });
 
-    if (user?.id !== args.data.userId) {
+    if (args.data.guildId === undefined && user?.id !== args.data.userId) {
```
alexkeating, line 76: Is this logic correct?  Shouldn't this be guildId not undefined and context user id does not equal args userId?

apps/protocol-api/src/prisma/resolvers/UserGuild.ts
```diff
@@ -54,19 +67,69 @@ export class GuildUserCustomResolver {
   ) {
     const address = req.session.siwe.data.address;
 
+    console.log('args', args);
```
alexkeating, line 70: This should probably be removed

apps/protocol-frontend/src/contexts/UserContext.tsx
```diff
@@ -230,6 +230,12 @@ export const UserContextProvider: React.FC<UserContextProps> = ({
     }
   }, [userDataByAddress, isAuthenticated]);
 
+  // useEffect(() => {
+  //   if (userDataByAddress && isAuthenticated) {
+  //     getUser();
+  //   }
+  // }, [userData?.guild_users, isAuthenticated, userDataByAddress]);
+
```
alexkeating, line 238: This should probably be commented back in

jonathanprozzi: Do we want to add this back in now or wait until we have the more robust UserContext refactor and use that instead?

apps/protocol-api/src/prisma/resolvers/UserGuild.ts
jonathanprozzi: Yup, great catch. I think when reworking this the other day I reversed it.

jonathanprozzi: Oh. I realize now why it's not this way -- we're creating the DAO first and then adding the users to it, so by the time it gets to this point it has an ID. Let me try to rework this.

We had to create the DAO first and then link the users to it otherwise it was creating a new DAO for each user

jonathanprozzi: This should be fixed now

apps/protocol-frontend/src/contexts/UserContext.tsx
jonathanprozzi: Issue with this that I'm seeing -- I split this into it's own effect to avoid the circular dependencies,  but then we'd introduce another lint warning (and another useEffect)

We could do that or refactor to use react query for the user DAO's check

alexkeating: If it is commented out we want to remove it

jonathanprozzi: Should be gone, let me double check if it didn't make the commit for some reason

PR #392, opened 2023-01-30
Replace `useMemo` hooks with table accessor functions

amrro:
## Linear Ticket

[ENG-846 - Three remaining tables to migrate](https://linear.app/govrn/issue/ENG-846/three-remaining-tables-to-migrate)

## Description

Replacing `useMemo` hooks with react-table accessor keys & functions. Besides, there couple of bugs: 
- In `AttestationTableType`, date of submission was used instead of date of engagement.
- In `MyAttestationsTableType`, for date of submission, first date in the array is used. I changed that to the the user's attestation date instead.


apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -148,36 +52,117 @@ const AttestationsTable = ({
...
   const deselectAll = () => {
     setRowSelection({});
   };
 
-  const data = useMemo<AttestationTableType[]>(
-    () =>
-      contributionsData.map(contribution => ({
-        id: contribution.id,
-        date_of_submission: contribution.date_of_submission,
-        date_of_engagement: contribution.date_of_submission,
```
amrro, line 164: date of submission was used instead of date of engagement

PR #397, opened 2023-02-01
dao-memberships-improvements-from-qa-2

jonathanprozzi:
## Linear Ticket

- [ENG-880](https://linear.app/govrn/issue/ENG-880/fixes-from-qa-session-2)

## Description

- This addresses several of the items from QA session 2:
  - Change name on DAO Settings page to {DAO Name} - Settings
  - Fix settings and profile links on DAO Cards (use DAO's id not the GuildUser id)
  - Hide import history table until we support other import methods
  - Add max height / move admins button for the DAO Members Set Admin table
  - Add 'Role' column to member addresses table
  - Restrict imports to less than 1000 at a time
  - Put admins to top in member addresses table (default sort - admins at top)
  - Pin user's address at top/distinguish it — need to think about this one a bit. Easy win is to use bold/gradient for this one
    - Using bolded for now, not pinned to top 
  - Change the DAO name placeholder on the manual DAO create name input

Remaining items are unchecked in the Linear ticket

## Screencaptures

https://user-images.githubusercontent.com/9438776/216163935-18a3379a-6bfb-4178-9b77-c112470b6722.mp4

^ temporarily removed the validation to test the table height, it's back in now

### DIFF
--- a/apps/protocol-frontend/src/components/CreateDaoLayout.tsx
+++ b/apps/protocol-frontend/src/components/CreateDaoLayout.tsx
@@ -4,7 +4,6 @@ import PageHeading from './PageHeading';
 import DaoTextareaForm from './DaoTextareaForm';
 import DaoImportCard from './DaoImportCard';
 import ModalWrapper from './ModalWrapper';
-import ImportHistoryTable from './ImportHistoryTable';
 
 const CreateDaoLayout = () => {
   // this will be removed once we add this to the dao settings page -- needed for testing the modal
@@ -35,7 +34,6 @@ const CreateDaoLayout = () => {
         width="100%"
       >
         <PageHeading>Create a DAO</PageHeading>
-        <ImportHistoryTable />
         <Grid
           padding={{ base: 4, lg: 0 }}
           templateColumns={{
--- a/apps/protocol-frontend/src/components/CreateDaoLayout.tsx
+++ b/apps/protocol-frontend/src/components/CreateDaoLayout.tsx
@@ -6,7 +6,6 @@ import DaoImportCard from './DaoImportCard';
 import ModalWrapper from './ModalWrapper';
 
 const CreateDaoLayout = () => {
-  // this will be removed once we add this to the dao settings page -- needed for testing the modal
   const { setModals } = useOverlay();
   const localOverlay = useOverlay();
 
@@ -14,7 +13,7 @@ const CreateDaoLayout = () => {
     setModals({ createDaoModal: true });
   };
 
-  // we can add the other methods in as they're ready
+  // we can add the other methods in as they're ready. we'll add the ImportHistoryTable back in at that point
   const daoImportMethods = [
     {
       importName: 'Create a DAO Manually',
--- a/apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
+++ b/apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
@@ -168,7 +168,7 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
         },
       },
     ];
-  }, []);
+  }, [userData?.address]);
 
   const table = useReactTable({
     data,
--- a/apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
+++ b/apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
@@ -100,7 +100,7 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
       {
         id: 'selection',
         header: 'Select Admin',
-        cell: ({ row, table }) => (
+        cell: ({ row }) => (
           <IndeterminateCheckbox
             {...{
               checked: row.getIsSelected(),
@@ -114,9 +114,8 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
         header: 'Member Address',
         accessorKey: 'user',
         cell: ({ getValue }: { getValue: Getter<UIGuildUsers['user']> }) => {
-          const { address } = getValue();
-          const memberAddress = getValue().address;
-          return address === userData?.address ? (
+          const value = getValue();
+          return value.address === userData?.address ? (
             <Text
               whiteSpace="normal"
               fontWeight="bold"
@@ -126,7 +125,7 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
               {getValue().address}
             </Text>
           ) : (
-            <Text whiteSpace="normal">{getValue().address}</Text>
+            <Text whiteSpace="normal">{value.address}</Text>
           );
         },
       },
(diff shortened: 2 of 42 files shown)


apps/protocol-frontend/src/components/CreateDaoLayout.tsx
```diff
@@ -4,18 +4,16 @@ import PageHeading from './PageHeading';
 import DaoTextareaForm from './DaoTextareaForm';
 import DaoImportCard from './DaoImportCard';
 import ModalWrapper from './ModalWrapper';
-import ImportHistoryTable from './ImportHistoryTable';
```
alexkeating, line 7: (arbitrary thread)

We should probably redirect somewhere when a dao is created. Especially if there is no table.

apps/protocol-frontend/src/components/CreateDaoLayout.tsx
```diff
@@ -35,7 +33,6 @@ const CreateDaoLayout = () => {
         width="100%"
       >
```
alexkeating, line 34: (aribtrary thread)

Tested "Change import status to Member not Recruit for users added directly via the Create a DAO or via DAO Settings" and my added user was still a recruit

apps/protocol-frontend/src/components/CreateDaoLayout.tsx
alexkeating: It may even make sense to move the modal directly to the create dao button in the profile

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
```diff
@@ -111,11 +114,61 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
...
         cell: ({ getValue }: { getValue: Getter<UIGuildUsers['user']> }) => {
-          return <Text whiteSpace="normal">{getValue().address}</Text>;
+          return getValue().address === userData?.address ? (
+            <Text
+              whiteSpace="normal"
+              fontWeight="bold"
+              bgGradient="linear(to-l, #7928CA, #FF0080)"
+              bgClip="text"
+            >
+              {getValue().address}
```
alexkeating, line 124: Let's move getValue to a variable to make this DRYer

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
```diff
@@ -111,11 +114,61 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
...
+      },
+      {
+        header: 'Role',
+        accessorKey: 'membershipStatus',
+        cell: ({
+          getValue,
+        }: {
+          getValue: Getter<UIGuildUsers['membershipStatus']>;
+        }) => {
+          if (getValue() === undefined || getValue() === null) return;
```
alexkeating, line 139: Same as above move to variable

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
```diff
@@ -111,11 +114,61 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
...
+                getValue()?.name === 'Admin'
+                  ? 'brand.purple'
+                  : getValue()?.name === 'Member'
+                  ? 'brand.magenta'
+                  : 'gray.200'
+              }
+              color={
+                getValue()?.name === 'Admin'
+                  ? 'white'
+                  : getValue()?.name === 'Member'
```
alexkeating, line 154: These don't need the ? because of the if statement above

apps/protocol-frontend/src/components/ImportHistoryTable.tsx
```diff
@@ -73,7 +73,7 @@ const ImportHistoryTable = () => {
         }) => {
           return (
             row.original.import_status.name === 'Completed' &&
-            userDaos?.get(getValue())?.membershipStatus?.name === 'admin' && (
+            userDaos?.get(getValue())?.membershipStatus?.name === 'Admin' && (
```
alexkeating, line 76: We should move these to an enum. This can be done in a separate ticket.

apps/protocol-frontend/src/components/CreateDaoLayout.tsx
alexkeating: Looks like #398 fixes this, please change the description as it says it fixes it here.

apps/protocol-frontend/src/components/CreateDaoLayout.tsx
jonathanprozzi: Good idea! Instead of linking to it from there, we should pop that open itself since it is a modal

apps/protocol-frontend/src/components/CreateDaoLayout.tsx
jonathanprozzi: Ah, my mistake. I copied that over from the ticket and forgot to remove. Thanks!

apps/protocol-frontend/src/components/ImportHistoryTable.tsx
jonathanprozzi: Totally agree. Especially so that it prevents mistakes like we see here ('admin' -> 'Admin')

apps/protocol-frontend/src/components/CreateDaoLayout.tsx
jonathanprozzi: On the redirect: I agree. I think once the context refactor is in we should redirect to the newly created DAO page. I held off on that since the user would need to refresh that page for them to be considered a member, but if we handle that via invalidation we can direct to that new page.

jonathanprozzi: One consideration with the modal -- that'll work for now, but when we have multiple methods we'd probably want to link to the Create Page (as it is now) where a user can choose import methods.

So if we introduce the modal now in the short term, we may want to change it once we have more methods. Wonder if that'll be less ideal than having it be the page link now since we'll end up there in the medium/longer term. 

@twblack88 curious for your thoughts here since I think there's pros/cons to both

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
jonathanprozzi: I thought so as well since we have that check but removing the optional chaining does cause an issue about getValue().name possibly being undefined or null.

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
jonathanprozzi: Pretty minor question, but what do we think about destructuring this vs. using a var:

`const { address } = getValue()`
vs. 
`const memberAddress = getValue().address`

Destructuring requires us to use the value as it exists (ie. `address`) whereas the variable lets us create a more expressive name, such as `memberAddress`

Thoughts @alexkeating @amrro

alexkeating: I was thinking more `const value = getValue()`. I am find with either.

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
alexkeating: I am guessing it is because the getValue function is being called under the if statement. My guess is the typing would remove the need for optional chaining if it was abstract to a variable and accessed via dot notation

jonathanprozzi: That's a good idea. I think so too, since it seemed that the function call was ignoring the type check.

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
jonathanprozzi: Doing the `const value = getValue()` and using dot notation worked with the typeguard.

PR #398, opened 2023-02-02
qa/change-import-member-status (ENG-887)

jonathanprozzi:
## Linear Ticket

- [ENG-887](https://linear.app/govrn/issue/ENG-887/change-import-status-to-member-not-recruit-for-users-added-directly)

## Description

- When users join a DAO via the 'Join a DAO' on profile they're added as a `Recruit`
- When users are added via the manual process, or added via the settings page, they're added as a `Member`
- Generalized the 

## Screencaptures

https://user-images.githubusercontent.com/9438776/216232618-3ac1dcd6-95c5-4bba-b6a2-9ff6f0955385.mp4

### DIFF
--- a/apps/protocol-frontend/src/hooks/useDaoUserCreate.ts
+++ b/apps/protocol-frontend/src/hooks/useDaoUserCreate.ts
@@ -9,7 +9,6 @@ interface DaoUserCreateProps {
 }
 
 export const useDaoUserCreate = () => {
-  console.log('mutation');
   const toast = useGovrnToast();
   const { govrnProtocol: govrn } = useUser();
   const queryClient = useQueryClient();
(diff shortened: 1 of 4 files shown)


apps/protocol-frontend/src/hooks/useDaoUserCreate.ts
```diff
@@ -43,7 +43,11 @@ export const useDaoUserCreate = () => {
             description: `${
               creatingNewDao === true
                 ? 'You have successfully created a new DAO.'
-                : 'You have successfully joined the DAO as a recruit.'
+                : `Successfully added to the DAO as a ${
+                    newDaoUser.membershipStatus
+                      ? newDaoUser.membershipStatus.toLowerCase()
+                      : 'recruit'
+                  }.`
```
alexkeating, line 50: This is a bit complex it will be clearer if this turnary is abstracted to a variable

alexkeating (commented):
Looks good and working, would be good to simplify the logic

apps/protocol-frontend/src/hooks/useDaoUserCreate.ts
jonathanprozzi: Great callout!

alexkeating (approved):

PR #400, opened 2023-02-02
Refactor User context

alexkeating:
## Description

- Convert user context to hooks to make invalidation easier

### DIFF
--- a/apps/protocol-frontend/project.json
+++ b/apps/protocol-frontend/project.json
@@ -63,7 +63,7 @@
       "outputs": ["{options.outputFile}"],
       "options": {
         "lintFilePatterns": ["apps/protocol-frontend/**/*.{ts,tsx,js,jsx}"],
-        "maxWarnings": 11
+        "maxWarnings": 10
       }
     },
     "test": {
--- a/apps/protocol-frontend/src/Routes.tsx
+++ b/apps/protocol-frontend/src/Routes.tsx
@@ -49,7 +49,7 @@ const RequireDaoUser = ({ children }: { children: JSX.Element }) => {
   const { userData } = useUser();
   const { data } = useUserGet({ userId: userData?.id });
   if (guildId) {
-    if (data?.userDaos.has(parseInt(guildId)) === false) {
+    if (data?.userDaos && data?.userDaos.has(parseInt(guildId)) === false) {
       return (
         <ErrorView
           errorMessage="You have to be a member of this DAO to view the DAO Dashboard."
--- a/apps/protocol-frontend/src/Routes.tsx
+++ b/apps/protocol-frontend/src/Routes.tsx
@@ -22,6 +22,7 @@ import RedirectHome from './pages/Redirect';
 import ContributionDetail from './pages/ContributionDetail';
 import CreateDao from './pages/CreateDao';
 import ErrorView from './components/ErrorView';
+import useUserGet from './hooks/useUserGet';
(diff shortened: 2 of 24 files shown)


apps/protocol-frontend/project.json
```diff
@@ -63,7 +63,7 @@
       "outputs": ["{options.outputFile}"],
       "options": {
         "lintFilePatterns": ["apps/protocol-frontend/**/*.{ts,tsx,js,jsx}"],
-        "maxWarnings": 11
+        "maxWarnings": 10
```
jonathanprozzi, line 66: 🙌

apps/protocol-frontend/src/Routes.tsx
```diff
@@ -44,10 +45,11 @@ const RequireActiveUser = ({ children }: { children: JSX.Element }) => {
 };
 
 const RequireDaoUser = ({ children }: { children: JSX.Element }) => {
-  const { isUserDaoMember } = useUser();
   const { guildId } = useParams();
+  const { userData } = useUser();
+  const { data } = useUserGet({ userId: userData?.id });
```
jonathanprozzi, line 50: Great idea having the hook call within the RequireDaoUser so that we're only making this query when a user is navigating to a Route wrapped with this. One worry I had was that we'd be making this call on every route change, but the wrapping avoids that.

apps/protocol-frontend/src/components/AttestationsTableShell.tsx
```diff
@@ -33,7 +36,7 @@ const AttestationsTableShell = () => {
       guilds: {
         some: {
           guild_id: {
-            in: userData?.guild_users.map(guild => guild.guild_id) || [],
+            in: guildIds,
```
jonathanprozzi, line 39: I like this, it's a much cleaner and readable way to have this.  I see these suggestions in your PR review that we should pull out this type of logic into variables so that it's more readable. We should include this in our best practices at this point 🙌

apps/protocol-frontend/src/hooks/useDaosList.ts
```diff
@@ -7,6 +7,9 @@ export const useDaosList = ({ ...args }) => {
   const { isLoading, isFetching, isError, error, data } = useQuery(
     ['userDaos', args],
     async () => {
+      if (!args.userId) {
```
jonathanprozzi, line 10: Good idea to include this here

jonathanprozzi (commented):
This looks great. Doing a bit more testing locally, but incredible improvements.

One thing I'm noticing: 
- We may want to standardize our export method for our hooks. Some are using named exports and some are using default

alexkeating:
> This looks great. Doing a bit more testing locally, but incredible improvements.
> 
> One thing I'm noticing:
> 
>     * We may want to standardize our export method for our hooks. Some are using named exports and some are using default

Do we have multiple hooks in a file? If not I think default makes sense

jonathanprozzi:
In testing, noticing that my DAOs dropdown for Creating and Editing contributions is empty.
image

We're using `userData?.id` as an arg in the hook there that populates it. Should be updated to use the new logic like you added in the Sidebar: ` {userData?.guild_users && userData?.guild_users.length > 0 ? (`

jonathanprozzi:
> 

I think that's a great deciding factor. Sometimes we have more than 1 with lists and infinite lists, but otherwise we don't. 
So we use named when more than 1, default otherwise

PR #409, opened 2023-02-07
Update profile dropdown

alexkeating:
## Description

- Add validated daos to the join dropdown


jonathanprozzi (approved):
Looks good! 
🎉 will be a great UX improvement for the Join DAO flow

PR #414, opened 2023-02-09
feature/improve-member-table-address-ux (ENG-917)

jonathanprozzi:
## Linear Ticket

https://linear.app/govrn/issue/ENG-917/improve-member-table-address-ux

## Description

This is the _option 2_ from our discussion. We can adjust the logic for this further if we'd like, but here's how it displays:

- Uses `name` or `display_name` if they exist (added `display_name` to our GuildUser fragment)
- If they don't exist, use the `address`
- If showing the name or display name, show the address on tooltip
- Only show the address tooltip if _not_ the connected user (figure they'd know their own address, but open to feedback there)
- I also made it so that we only show the tooltip on addresses that aren't the connected user. I don't necessarily think they'd need to see a tooltip for their own address.

Additionally, only showing the tooltip if the user's address isn't showing (so if we're showing the name/display name, then the address shows on hover)

## Screencaptures

https://user-images.githubusercontent.com/9438776/217836658-a1253f03-ae69-4e60-a44e-b09c8ef1a8e3.mp4

### DIFF
--- a/apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
+++ b/apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
@@ -28,6 +28,7 @@ import { useDaoUsersInfiniteList } from '../hooks/useDaoUsersList';
 import { RowSelectionState } from '@tanstack/table-core';
 import { SortOrder } from '@govrn/protocol-client';
 import GovrnTable from './GovrnTable';
+import { displayAddress } from '../utils/web3';
 
 interface DaoSettingsMembersTableProps {
   daoId: number;
@@ -111,6 +112,7 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
         accessorKey: 'user',
         cell: ({ getValue }: { getValue: Getter<UIGuildUsers['user']> }) => {
           const value = getValue();
+          const displayMemberName = value.name || displayAddress(value.address);
           return value.address === userData?.address ? (
             <Tooltip
               variant="primary"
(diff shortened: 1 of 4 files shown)


apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
```diff
@@ -48,6 +57,8 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
     );
   }, [daoUsersData]);
 
+  console.log('data', data);
```
alexkeating, line 60: Rouge log

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
```diff
@@ -97,21 +108,32 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
...
             <Text
               whiteSpace="normal"
               fontWeight="bold"
               bgGradient="linear(to-l, #7928CA, #FF0080)"
               bgClip="text"
             >
-              {value.address}
+              {displayMemberName}
             </Text>
+          ) : value.name || value.display_name ? (
```
alexkeating, line 126: Let's abstract this turnary, outside of the jsx

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
```diff
@@ -97,21 +108,32 @@ const DaoSettingsMembersTable = ({ daoId }: DaoSettingsMembersTableProps) => {
...
+            <Tooltip
+              variant="primary"
+              label={displayAddress(value.address)}
```
alexkeating, line 129: Shouldn't this be the full address

alexkeating: Isn't the goal here to verify the address

alexkeating (commented):
Really close!

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
jonathanprozzi: ha thanks, good catch

apps/protocol-frontend/src/components/DaoSettingsMembersTable.tsx
jonathanprozzi: That's what I thought too -- I think I asked this yesterday but maybe I wasn't clear. I agree that this should be full and if it's displayed in the table it's truncated.

jonathanprozzi: I removed the truncate here since I agree -- if it's too cumbersome we can revisit something like positioning the tooltip

PR #421, opened 2023-02-15
fix/redirect-flow-check (ENG-939) -- Hold on merge

jonathanprozzi:
## Linear Ticket

- [ENG-939](https://linear.app/govrn/issue/ENG-939/bug-user-can-interrupt-auth-flow-check)

## Description

- See ticket for details
- If we'd like, we can add in a specific Error view similar to the DAO users check. I opted for a redirect for now

### DIFF
--- a/apps/protocol-frontend/src/Routes.tsx
+++ b/apps/protocol-frontend/src/Routes.tsx
@@ -53,19 +53,6 @@ const RequireActiveUser = ({ children }: { children: JSX.Element }) => {
   return children;
 };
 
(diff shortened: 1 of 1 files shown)


jonathanprozzi:
Noticing an issue with this when testing more -- refreshing takes back to the /

jonathanprozzi:
Fixing this from our conversation @alexkeating

alexkeating:
Should this pr be closed then?

jonathanprozzi:
Let's close this and I'll reopen one when the new, more robust flow is in place.

PR #422, opened 2023-02-15
Update bot flow to send link and helper message

amrro:
## Linear Ticket

- [ENG-895 - Update bot flow to send link and helper message](https://linear.app/govrn/issue/ENG-895/update-bot-flow-to-send-link-and-helper-message)

## Description
Update `/join` command to use the connect discord through the Govrn Web-app.


apps/kevin-malone/bot/common/threads/onboarding.py
```diff
@@ -73,7 +71,22 @@ async def control_hook(self, message, user_id):
...
+        return StepKeys.PROMPT_USER_CONNECT_WALLET.value
+
+
+class PromptUserToConnectWallet(BaseStep):
+    """Step to submit wallet address for the govrn profile"""
+
+    name = StepKeys.PROMPT_USER_CONNECT_WALLET.value
+
+    msg = (
+        f"It doesn't look like your wallet is connected to this discord account please connected them at this link {constants.Bot.govrn_discord_auth_url}.\n\n"
```
jonathanprozzi, line 83: The new `/signature` route that we'll have in place for supporting this flow takes in an optional query param for the display name as `displayName` so we could add that to this link as well, such as:

`/signature?displayName=km-user-flow` -- the Signature flow uses this display name (if it's there) when creating a new user

alexkeating:
I think this should be working now

apps/kevin-malone/bot/common/threads/onboarding.py
```diff
@@ -73,7 +70,33 @@ async def control_hook(self, message, user_id):
...
+        super().__init__()
+        self.cache = cache
+        self.guild_id = guild_id
+        self.bot = bot
+
+    async def send(self, message, user_id):
+        channel = message.channel
+        discord_user = await self.bot.fetch_user(user_id)
+        msg = (
+            "It doesn't look like your wallet"
```
Flip-Liquid, line 91: can we extract this copy in to config.py? There are a few examples in there of how to format so that we can insert the right link.

alexkeating: Yes, we can move the copy to the config.py

apps/kevin-malone/bot/common/threads/onboarding.py
```diff
@@ -73,7 +70,33 @@ async def control_hook(self, message, user_id):
...
+            " is connected to this discord account please "
```
Flip-Liquid, line 92: In this case, we don't know if the user already has an account setup with Govrn, but has not yet linked their discord, or if this is their first time actually joining, right? Can we update the copy to the following to reflect instructions for a first time join? i.e.: "Hello! We don't have a Govrn profile associated with your discord quite yet. If you'd like to setup a Govrn profile, or associate your discord with an existing Govrn profile, you can do so here: {constants.Bot.frontend_url}/#/signature?displayName={discord_user.display_name}. After that, just run /join again to hop into your guild!"

Flip-Liquid (changes requested):
just extract + minor edits to the copy, lgtm otherwise tho. have you tested this locally @amrro ?

apps/kevin-malone/bot/common/threads/onboarding.py
alexkeating: Doing now

apps/kevin-malone/bot/common/threads/onboarding.py
alexkeating: Going to keep this here as it uses the display name variable. It seems a bit premature to abstract now

PR #423, opened 2023-02-15
feature/join-flow-page (ENG-894)

jonathanprozzi:
## Linear Ticket

- [ENG-894](https://linear.app/govrn/issue/ENG-894/implement-standalone-page)

## Description

- Add frontend support for the new KM join flow
- Currently creating a new user if they don't already exist is called when connecting to Discord, but thinking this should be after the signature instead per our conversation @alexkeating 
  - We can test the rest of the flow while I change this ^

## Screencaptures

1) Successful flow for an existing user:

https://user-images.githubusercontent.com/9438776/219101137-1b6aafde-b6f6-4f98-87b9-a11d5f3f84a9.mp4

2) New user with display name coming from KM:

https://user-images.githubusercontent.com/9438776/219101177-1d2be8d9-8edd-4a7d-a379-06c8fe758e69.mp4

3) New user without display name coming from KM:

https://user-images.githubusercontent.com/9438776/219101190-ec1194d9-c6b3-4509-94b7-41d894b1a52b.mp4


apps/protocol-frontend/src/components/SignatureLayout.tsx
```diff
@@ -0,0 +1,278 @@
...
+            {isConnected && !isDiscordConnected(userData) ? ( // stage 2: discord oauth
+              userData ? (
+                <StageTwoExistingUser />
+              ) : (
+                <StageTwoNewUser />
+              )
+            ) : null}
+            {isConnected && userData && isDiscordConnected(userData) ? ( // stage 3: success and next steps
+              <StageThreeSuccess />
+            ) : null}
```
alexkeating, line 270: We should move some of this outside of the JSX and possibly encapsulated in a component rather than having these nested ternaries.

apps/protocol-frontend/src/hooks/useUserCreate.ts
```diff
@@ -4,30 +4,39 @@ import { useMutation, useQueryClient } from '@tanstack/react-query';
...
+        // only show toast if specified
+        if (showToast === true) {
+          toast.success({
+            title: 'User Created',
+            description: `Your username has been created with your address: ${address}. Let's report your first contribution!`,
+          });
+        }
       },
       onError: error => {
+        // always show error toast
```
alexkeating, line 39: Extraneous comments

apps/protocol-frontend/src/pages/Signature.tsx
```diff
@@ -0,0 +1,12 @@
+import { Box } from '@chakra-ui/react';
+import SignatureLayout from '../components/SignatureLayout';
+
+const Signature = () => {
```
alexkeating, line 4: We should definitely rename this to something a little more specific to this flow.

apps/protocol-frontend/src/components/SignatureLayout.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const { isConnected, address } = useAccount();
+  const [searchParams] = useSearchParams();
+  const displayNameFromParams = searchParams.get('displayName');
+  const { mutateAsync: createUser } = useUserCreate();
+  const { userDataByAddress, userData } = useUser();
+  const { displayName } = useDisplayName();
+
+  const BACKEND_ADDR = `${BASE_URL}`;
+  const DISCORD_CLIENT_ID = import.meta.env.VITE_DISCORD_CLIENT_ID;
+  const DISCORD_REDIRECT_URI = import.meta.env.VITE_DISCORD_REDIRECT_URI;
```
alexkeating, line 27: These would be good to add to the docs

alexkeating (commented):
Overall, looks good and works well

apps/protocol-frontend/src/components/SignatureLayout.tsx
jonathanprozzi: I agree. I want to revisit that - it's how I did it initially and I ran into some strange re renders so I moved to these.

apps/protocol-frontend/src/pages/Signature.tsx
jonathanprozzi: Good call. SignatureJoinFlow? SignatureDiscordFlow ? Something like that

apps/protocol-frontend/src/components/SignatureLayout.tsx
jonathanprozzi: Good call, will add 👍

apps/protocol-frontend/src/pages/Signature.tsx
jonathanprozzi: Went with DiscordSignature and DiscordSignatureLayout for now. We can chat about this briefly at standup but agree should be clearer

PR #427, opened 2023-02-22
feature/cta-component (ENG-954)

jonathanprozzi:
## Linear Ticket

- [ENG-954](https://linear.app/govrn/issue/ENG-954/cta-component)

## Description

I originally included this in a branch with the other empty state changes (contributions) but then realized that it was cleaner to split this off first and put it in its own PR since I was touching other logic.

You can run our Storybook locally with `nx run protocol-ui:storybook` to test locally. I included these in the short video included in the PR.

The component takes in a heading, emoji, copy, and children. 

- Added the GovrnCta component to our Storybook
- Fixed a few things in Storybook -- it wasn't properly pulling in our themes and variants

## Screencaptures

https://user-images.githubusercontent.com/9438776/220701702-5a9e8984-c4d0-4487-9713-df16193b09e9.mp4

### DIFF
--- a/libs/protocol-ui/src/components/molecules/GovrnCta/GovrnCta.tsx
+++ b/libs/protocol-ui/src/components/molecules/GovrnCta/GovrnCta.tsx
@@ -1,0 +1,60 @@
+import { Heading, Divider, Flex, Text } from '@chakra-ui/react';
+
+export interface GovrnCtaProps {
+  heading: string;
+  emoji?: string;
+  copy?: string;
+  children?: React.ReactNode;
+}
+
+const GovrnCta = ({ heading, emoji, copy, children }: GovrnCtaProps) => (
+  <Flex
+    direction="column"
+    alignItems="center"
+    justifyContent="center"
+    background="white"
+    boxShadow="sm"
+    borderRadius={{ base: 'none', md: 'lg' }}
+  >
+    <Flex direction="row" alignItems="baseline" fontSize="3xl" paddingY={8}>
+      {emoji ? (
+        <Text
+          as="span"
+          role="img"
+          aria-labelledby="eye emoji for the CTA"
+          fontSize="inherit"
+        >
+          {emoji}
+        </Text>
+      ) : null}
+      <Heading as="h3" fontSize="3xl" fontWeight="600">
+        {heading}
+      </Heading>
+    </Flex>
+    <Text paddingBottom={8} fontSize="md">
+      {copy}
+    </Text>
+    {children ? (
+      <Flex
+        direction="column"
+        justifyContent="center"
+        alignItems="center"
+        paddingX={8}
+        width="100%"
+      >
+        <Divider color="gray.200" />
+        <Flex
+          direction="row"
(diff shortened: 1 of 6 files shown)


libs/protocol-ui/src/components/molecules/GovrnCta/GovrnCta.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {emoji ? (
+        <Text
+          as="span"
+          role="img"
+          aria-labelledby="eye emoji for the CTA"
+          fontSize="inherit"
+        >
+          {emoji}
+        </Text>
+      ) : null}
```
alexkeating, line 29: Why not use &&? It is more concise

libs/protocol-ui/src/components/molecules/GovrnCta/GovrnCta.tsx
```diff
@@ -0,0 +1,60 @@
...
+          direction="row"
+          alignItems="center"
+          justifyContent="space-around"
+          paddingY={8}
+          width="100%"
+        >
+          {children}
+        </Flex>
+      </Flex>
+    ) : null}
```
alexkeating, line 56: Same here

libs/protocol-ui/src/stories/molecules/GovrnCta.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const TwoButtons = (args: GovrnCtaProps) => (
+  <Box backgroundColor="gray.50">
+    <Box padding={10}>
+      <GovrnCta {...sharedArgs} children={Buttons.slice(0, 2)} />
+    </Box>
+  </Box>
+);
+
+export const ThreeButtons = (args: GovrnCtaProps) => (
```
alexkeating, line 40: These seem a bit excessive. We should probably only have one component and pass in the appropriate buttons where they are needed. This will give us more flexibility

jonathanprozzi: Oh these are just demos for the story. It takes in a child which is however many elements it needs.

jonathanprozzi: Actual component only takes in children - this shows how it looks with 0-3

alexkeating: Ah okay cool

libs/protocol-ui/src/components/molecules/GovrnCta/GovrnCta.tsx
jonathanprozzi: Mainly to avoid potential issues if children evaluates to 0 such as if it's coming in as an empty array. It'll render as a "0"

alexkeating: emoji won't have that problem right? And even for children it is probably preferable to move the turnary out of the JSX

PR #428, opened 2023-02-23
Remove global logger

alexkeating:
## Description

- Remove global error toast

## Ticket 
- https://linear.app/govrn/issue/ENG-924/error-toasts-when-signing-in-to-govrn-app-w-wallet


jonathanprozzi (approved):
Good call. We don't really need this any more anyway since we have more specific error messages in our hooks.

PR #429, opened 2023-02-23
On mint save date of engagement not date of submission

alexkeating:
## Linear Ticket
- https://linear.app/govrn/issue/ENG-964/change-date-in-minted-view

## Description

- Looks like we are saving on submission during a mint. We will need to pair this with updating the backfilled data.


jonathanprozzi (approved):
This makes more sense to show over the submission date. Good idea

PR #432, opened 2023-02-24
Non nullable membership

amrro:
## Linear Ticket
- [ENG-940 - Create database constraint so a status is assigned
](https://linear.app/govrn/issue/ENG-940/create-database-constraint-so-a-status-is-assigned)

## Description

Update `GuildUser` to have non-optional FK to `GuildMembershipStatus`, linking null fk for existing records to be `Member` value.

### DIFF
--- a/apps/protocol-api/project.json
+++ b/apps/protocol-api/project.json
@@ -37,7 +37,7 @@
       "outputs": ["{options.outputFile}"],
       "options": {
         "lintFilePatterns": ["apps/protocol-api/**/*.ts"],
-        "maxWarnings": 6
+        "maxWarnings": 3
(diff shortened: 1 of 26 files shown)


apps/protocol-api/project.json
```diff
@@ -37,7 +37,7 @@
       "outputs": ["{options.outputFile}"],
       "options": {
         "lintFilePatterns": ["apps/protocol-api/**/*.ts"],
-        "maxWarnings": 6
+        "maxWarnings": 3
```
alexkeating, line 40: Good job!

alexkeating (approved):

PR #434, opened 2023-02-24
contribution-table-enhancements

jonathanprozzi:
## Linear Ticket

- [ENG-985](https://linear.app/govrn/issue/ENG-958/add-address-on-hover-on-the-contributor)
- [ENG-813](https://linear.app/govrn/issue/ENG-813/remove-dao-column-contribution-types-page)

## Description

- Shows address on hover on Attestations, My Attestations, and Recent Contributions (DAO Dashboard) tables. This only shows when the contributor has a display name set (similar to the DAO Settings Member table)
- Removed the 'DAO' column and name from the Contribution Types table
- Added a link to Contribution Details via the _Last Occurrence_ column

## Screencaptures

Address on hover:

address-hover-attestations
address-hover-recent-contributions

Contribution Types table:

contribution-types-link

### DIFF
--- a/apps/protocol-frontend/src/components/AttestationsTable.tsx
+++ b/apps/protocol-frontend/src/components/AttestationsTable.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
-import { Box, Flex, Stack, Button, Text, Tooltip } from '@chakra-ui/react';
+import { Box, Flex, Stack, Button, Text } from '@chakra-ui/react';
 import { Link } from 'react-router-dom';
 import {
   ColumnDef,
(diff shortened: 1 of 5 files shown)


apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -165,6 +184,7 @@ const AttestationsTable = ({
     getSortedRowModel: getSortedRowModel(),
     getFilteredRowModel: getFilteredRowModel(),
   });
+  console.log('table data', data);
```
alexkeating, line 187: Excess console log

apps/protocol-frontend/src/components/RecentContributionsTable.tsx
```diff
@@ -74,9 +75,26 @@ const columnsDef: ColumnDef<UIContribution>[] = [
...
+          variant="primary"
+          label={value.address}
+          fontSize="sm"
+          placement="right"
+        >
+          <Text whiteSpace="normal">{displayMemberName}</Text>
+        </Tooltip>
+      ) : (
+        <Text whiteSpace="normal">{displayMemberName}</Text>
+      );
```
alexkeating, line 97: Should we abstract this logic into a component because it is being used in three places?

jonathanprozzi: Good idea, I think it's time to do that at this point 👍 
Had a similar thought about the emojis that we're using in the CTA and Showcase (basically creating an emoji component similar to how Icon components work -- it can handle the span and aria-labelledby)

jonathanprozzi: Created a `MemberDisplayName` component that handles the logic and display for this now.

We have 1 more instance of a _similar_ logic that has additional rules (in the Member Settings table) so I left that one for now, but this replaces the logic in the Attestations/My Attestations/Recent Contributions tables.

PR #439, opened 2023-03-01
Improve Select components' type definitions & support Grouping

amrro:
## Linear Ticket

- Related to: [ENG-546 - Organize taxonomy dropdown by DAO on the reporting form](https://linear.app/govrn/issue/ENG-546/organize-taxonomy-dropdown-by-dao-on-the-reporting-form)

## Description

- Introduce common types for Select components: `SelectOption`, group types, and Common props type `BaseSelectProps`.
- Add support for Groups in Select Dropdowns.
- Set Options' properties `label` & `value` as non-optional and ensure that value can be only a `string`, `number` or both if necessary.
- Document the purpose of each prop to provide better understanding and clarity.
- Move all Select components under the Select directory for better organization and accessibility.


libs/protocol-ui/src/components/molecules/Select/ControlledSelect.tsx
```diff
@@ -1,47 +1,27 @@
 import React from 'react';
 import ReactSelect from 'react-select';
 import { Stack, Box, Text } from '@chakra-ui/react';
+import { BaseSelectProps, ValueType } from './types';
```
jonathanprozzi, line 4: Great idea, streamlines it by quite a lot

libs/protocol-ui/src/components/molecules/Select/types.ts
```diff
@@ -0,0 +1,79 @@
+import { GroupBase, OptionsOrGroups } from 'react-select';
```
jonathanprozzi, line 1: This is a great idea. Nice improvement!

apps/protocol-frontend/src/utils/constants.ts
```diff
@@ -44,11 +44,11 @@ export const WEEK = 1;
 export const MONTH = 4;
 export const QUARTER = 12;
 export const YEAR = 52;
-export const DEFAULT_DATE_RANGES = [
-  { value: WEEK, label: 'Last Week' },
-  { value: MONTH, label: 'Last Month' },
-  { value: QUARTER, label: 'Last Quarter' },
-  { value: YEAR, label: 'Last Year' },
+export const DEFAULT_DATE_RANGES: SelectOption<number>[] = [
+  { label: 'Last Week', value: WEEK },
```
jonathanprozzi, line 48: Great improvement

jonathanprozzi (commented):
Great improvements and streamlining! I agree with moving these all into the Select folder as well.

Going and doing a quick test of the Selects but this is a great addition / refactor 🔥

amrro:
Good Catch @jonathanprozzi!

It seems odd that although the `setValue` method is being called from the local form, the values being passed to submit include a `daoId` is null.

jonathanprozzi:
> Good Catch @jonathanprozzi!
> 
> It seems odd that although the `setValue` method is being called from the DAOs select, the values being passed to submit include a `daoId` is null.
> 
> **Update**: Fixed [c0a0319](https://github.com/Govrn-HQ/govrn-monorepo/commit/c0a0319c7c5dd507b8ec015ccfc4e5e828d8398d). was `useEffect` deps.

That's what I thought too -- nice!

jonathanprozzi:
Nice, that fixed the issue.

I'm getting an error on the Edit mutation when setting a DAO to 'No DAO' but not sure it relates to this PR. Going to identify the issue. The forms and selects look good!

apps/protocol-frontend/src/components/EditContributionForm.tsx
```diff
@@ -76,15 +77,15 @@ const EditContributionForm = ({ contribution }: EditContributionFormProps) => {
...
     useUserData?.guild_users.map(dao => ({
       value: dao.guild.id,
       label: dao.guild.name ?? '',
     })) || [];
 
-  const daoReset = [
+  const daoReset: Option<number>[] = [
     {
-      value: null,
+      value: -1,
```
jonathanprozzi, line 88: It looks like this change is breaking the mutation -- we'll likely need to update the resolver. Changing this back to `null` resolves the error I mentioned in another comment

jonathanprozzi: The resolver has a check for null -- if we want to use `-1` instead we'll want to modify the contribution update resolver. Or we can modify the `SelectOption` type to allow for `null` and keep this as `null` on the reset.

@alexkeating what do you think?

alexkeating:
I think #440 handled this

PR #440, opened 2023-03-02
Group activity Types by associated DAO's name

amrro:
## Linear Ticket

[ENG-546 - Organize taxonomy dropdown by DAO on the reporting form](https://linear.app/govrn/issue/ENG-546/organize-taxonomy-dropdown-by-dao-on-the-reporting-form)

## Description

- This PR groups the activity types by the associated DAO's name in the Edit & Report Contribution forms. In doing so, we had to modify the `listActivityTypesByUser` query to include the metadata about the guild associated with each activity type. If there is a better way to accomplish this, please advise.

- To improve performance, some mappings are wrapped in a `useMemo` hook.
- To utilize Groups from `react-select` library, this PR benefits from changes PR #439. I think its better to review 439 & merge first to avoid double review.

Generally speaking, this change enhance the user experience and make it easier to navigate and select the relevant activity types in the Edit  & Report Contribution forms.


alexkeating:
@jonathanprozzi If you can give this another look that would be great

jonathanprozzi:
The dropdown looks great and this is a massive UX improvement! 

A few small things:

- Wondering if we'd want to add a 'Default' or 'User Generated' type of group for the ones that aren't from a user's DAOs sot that it also has a group header?
- Noticing that when I choose an activity type from a DAO that there is an error where it's passing in the whole object:
image

apps/protocol-api/src/main.ts
```diff
@@ -231,6 +232,11 @@ const permissions = shield(
       contribution_reporting_channel: or(hasToken, isAuthenticated),
       status: or(hasToken, isAuthenticated),
     },
+    GuildActivityType: {
```
alexkeating, line 235: (Arbitrary comment area)

Yes, we should add a personal section, but I think we can do this in a fast follow ticket. As a user can still add custom types, and I think the personal will require a whole new back to front endpoint.

jonathanprozzi (approved):
Solid!
Stray console log but all good!

PR #446, opened 2023-03-08
feature/verification-framework-dao-form (ENG-979)

jonathanprozzi:
## Linear Ticket

- [ENG-979](https://linear.app/govrn/issue/ENG-979/create-verification-in-dao-form)
- [ENG-984](https://linear.app/govrn/issue/ENG-984/create-numberinput-component-and-add-to-ui-library)

## Description

- This adds the Verification Framework section, the Verification Framework Form, and a NumberInput component (to our UI library)
- Basic validation has been added, but we'll likely fine tune all of this (validation, min/max number, etc.) when we wire it up to the backend [ENG-980](https://linear.app/govrn/issue/ENG-980/hook-up-verification-form-to-the-backend) 
- Right now, both fields are required and the number has a min / max of 1 and 10 as an example. We can remove the min/max (or keep the min, remove the max, etc.) based on how we set up the backend.

## Screencaptures

vf-form-ui

### DIFF
--- a/apps/protocol-frontend/src/components/DaoSettingsLayout.tsx
+++ b/apps/protocol-frontend/src/components/DaoSettingsLayout.tsx
@@ -9,6 +9,8 @@ import DaoNameUpdateForm from './DaoNameUpdateForm';
 import DaoSettingsMemberUpdateForm from './DaoSettingsMemberUpdateForm';
 import DaoSettingsMembersTable from './DaoSettingsMembersTable';
 import ErrorView from './ErrorView';
+import FeatureFlagWrapper from './FeatureFlagWrapper';
+import VerificationFramework from './VerificationFramework';
 
 const DaoSettingsLayout = () => {
   const { guildId } = useParams();
@@ -94,6 +96,9 @@ const DaoSettingsLayout = () => {
           <DaoSettingsMembersTable daoId={parseInt(guildId ? guildId : '')} />
         </Flex>
       </Flex>
+      <FeatureFlagWrapper>
+        <VerificationFramework />
+      </FeatureFlagWrapper>
     </Box>
   );
 };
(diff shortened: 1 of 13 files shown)


apps/protocol-frontend/src/components/DaoSettingsLayout.tsx
```diff
@@ -9,6 +9,8 @@ import DaoNameUpdateForm from './DaoNameUpdateForm';
 import DaoSettingsMemberUpdateForm from './DaoSettingsMemberUpdateForm';
 import DaoSettingsMembersTable from './DaoSettingsMembersTable';
 import ErrorView from './ErrorView';
+import FeatureFlagWrapper from './FeatureFlagWrapper';
+import VerificationFramework from './VerificationFramework';
```
alexkeating, line 14: (Arbitrary comment)

I think this was outside of the scopes of the ticket, but it should probably default to an empty state where framework is none and the number input is hidden

alexkeating (approved):

apps/protocol-frontend/src/components/DaoSettingsLayout.tsx
alexkeating: We can chose to do this in this pr or to create a new ticket

jonathanprozzi: Great call. I'll work on that tomorrow. Let's do it as a separate PR since I'll prob hack on a few designs

jonathanprozzi: Going to make another ticket for this 👍

jonathanprozzi:
Fixed conflicts 👍

jonathanprozzi:
Ah, looks like I need to fix a few things to work with the new Select types. Will clean that up and push up fix

PR #456, opened 2023-03-15
feature/verification-framework-empty-state (ENG-988)

jonathanprozzi:
## Linear Ticket

- [ENG-988](https://linear.app/govrn/issue/ENG-988/verification-framework-empty-state)

## Description

- Adds in a basic empty state if the selected verification framework is 'none'
- We can adjust the copy and tweak this once everything is fully wired up
- Thinking for the future we'll likely want to extend the conditional fields since other frameworks may need additional inputs (similar to how the # of attesters is revealed once that framework is chosen)

## Screencaptures

vf-empty-state


apps/protocol-frontend/src/components/VerificationFrameworkForm.tsx
```diff
@@ -63,15 +69,21 @@ const VerificationFrameworkForm = () => {
...
-          <NumberInput
-            name="numberOfAttestors"
-            label="Choose the Number of Attestors"
-            defaultValue={1}
-            min={1}
-            max={10}
-            localForm={localForm}
-          />
-
+          {watch('verificationFramework') === 'none' ? (
```
alexkeating, line 72: When I load the page, number of attestors is showing. It should by default be hidden unless number of attestors is selected.

alexkeating (commented):
One comment

apps/protocol-frontend/src/components/VerificationFrameworkForm.tsx
jonathanprozzi: oh, interesting! i'll be sure to set the default to 'none' -- let me test that again. thanks!

PR #460, opened 2023-03-15
Drop date of attestation

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/ENG-999/remove-date-of-attestation-column


jonathanprozzi:
This looks good but I think we need to remove the `date_of_attestation` from the Attestation resolver:
`date_of_attestation: new Date(Date.now()).toISOString(),` since it's no longer in the schema

PR #465, opened 2023-03-20
WIP add verification job scaffold

Flip-Liquid:
## Linear Ticket

## Description

## Screencaptures

### DIFF
--- a/apps/jobs/contribution-verification/jest.config.ts
+++ b/apps/jobs/contribution-verification/jest.config.ts
@@ -1,0 +1,16 @@
+/* eslint-disable */
+export default {
+  displayName: 'jobs-contribution-verification',
+  preset: '../../../jest.preset.js',
+  globals: {
+    'ts-jest': {
+      tsconfig: '<rootDir>/tsconfig.spec.json',
+    },
+  },
+  testEnvironment: 'node',
+  transform: {
+    '^.+\\.[tj]s$': 'ts-jest',
+  },
+  moduleFileExtensions: ['ts', 'js', 'html'],
+  coverageDirectory: '../../../coverage/apps/jobs/contribution-verification',
+};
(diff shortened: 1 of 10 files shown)


apps/jobs/contribution-verification/src/environments/environment.ts
```diff
@@ -0,0 +1,3 @@
+export const environment = {
+  production: false,
+};
```
alexkeating, line 3: These environment files can be removed

alexkeating (commented):
Code looks, good going to test locally and see if anything else pops up

apps/protocol-api/src/permissions/index.ts
```diff
@@ -213,6 +213,7 @@ const JOB_ONLY_QUERIES = {
   linearIssues: hasToken,
   contributionStatuses: hasToken,
   chain: hasToken,
+  guildContributions: hasToken,
```
alexkeating, line 216: We need to add the permission token to this file

alexkeating: It will also be good to add this to the readme

apps/jobs/contribution-verification/jest.config.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable */
```
alexkeating, line 1: (arbitrary comment)

We should add a readme for this job

apps/jobs/contribution-verification/src/db.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export async function getUnverifiedContributions(daoId: number) {
+  const result = await govrn.guildContribution.list({
+    where: {
+      guild_id: {
+        equals: daoId,
+      },
+      verificationStatus: {
+        is: {
+          name: { equals: 'unverified' },
```
alexkeating, line 40: This doesn't handle the null verification status

alexkeating: Also, I think these names need to be capitalized

alexkeating (approved):
This looks good, we will merge after the release

PR #470, opened 2023-03-22
Default to validated

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/ENG-1014/default-daos-to-validated


jonathanprozzi (approved):
Awesome. This means that any DAO that users create will automatically show up in the Join list now, right? 

Only concern is that the lists could get fairly long.

PR #472, opened 2023-03-24
Add user activities

alexkeating:
## Linear Ticket

https://linear.app/govrn/issue/ENG-998/add-user-activity-types-to-activity-type-dropdown


jonathanprozzi:
We need to update the `useContributionCreate` and `useContributionUpdate` invalidation to be this:
`        queryClient.invalidateQueries({ queryKey: ['userActivityList'] }); // invalidate the activity types `

It looks like we have another hook called `useUserActivityTypes` and we were using that key for invalidation instead of this one. Passing in with the queryKey in the params also invalidates this regardless of the long args being passed in (so we don't need to recreate the exact args in the invalidation) @alexkeating

alexkeating:
> We need to update the `useContributionCreate` and `useContributionUpdate` invalidation to be this: ` queryClient.invalidateQueries({ queryKey: ['userActivityList'] }); // invalidate the activity types`
> 
> It looks like we have another hook called `useUserActivityTypes` and we were using that key for invalidation instead of this one. Passing in with the queryKey in the params also invalidates this regardless of the long args being passed in (so we don't need to recreate the exact args in the invalidation) @alexkeating

This seems to be an existing bug

jonathanprozzi:
> > We need to update the `useContributionCreate` and `useContributionUpdate` invalidation to be this: ` queryClient.invalidateQueries({ queryKey: ['userActivityList'] }); // invalidate the activity types`
> > It looks like we have another hook called `useUserActivityTypes` and we were using that key for invalidation instead of this one. Passing in with the queryKey in the params also invalidates this regardless of the long args being passed in (so we don't need to recreate the exact args in the invalidation) @alexkeating
> 
> This seems to be an existing bug

We can merge this in and then reconcile that -- can add to the QA tickets or as another ticket 👍

PR #476, opened 2023-03-28
feature/verification-framework-add-dao-column-tables (ENG-978)

jonathanprozzi:
## Linear Ticket

- [ENG-978](https://linear.app/govrn/issue/ENG-978/add-pill-to-dao-column-in-attestation-table-and-minted-contributions)
- [ENG-976](https://linear.app/govrn/issue/ENG-976/create-different-tooltip-hover-states)

## Description

- This adds in the pills and hover states to the _Status_ and _DAO_ columns in the Attestations table and connects to the values coming from our API
- Note that the `attestation_threshold` may make this tricky to test at first since it's largely going to be null until we have the sync job and are setting the threshold on initial contribution creation (I'm working on this now but wanted to split this PR off)
  - Can test this locally by setting the # on each `GuildContribution` 

## Screencaptures

vf-attestation-table

If a contribution is verified for an attributed DAO, the pill in the DAO column will be purple with a checkmark like in the mock.


apps/protocol-frontend/src/components/AttestationsTable.tsx
```diff
@@ -110,21 +109,49 @@ const AttestationsTable = ({
...
-              >
-                {statusEmojiSelect(status)}
-              </span>{' '}
-            </Text>
+          const statusMapHover =
+            status === null
+              ? 'noFramework'
+              : status === 'Verified'
+              ? 'Verified'
+              : 'Unverified';
```
alexkeating, line 126: Chaining turnarys is kind of tough to read. An if/else statement may be more readable

apps/protocol-frontend/src/components/VerificationHover.tsx
```diff
@@ -0,0 +1,47 @@
...
+      label:
+        'Feel free to attest to this contribution! Even though it has been verified for this DAO, additional attestations is good. ',
+      variant: 'tertiary',
+    },
+    Unverified: {
+      label:
+        threshold === 1
+          ? `This contribution only needs ${threshold} more attestation to be verified by your DAO! Be the one to push it through!`
+          : `This contribution needs ${threshold} more attestations to be verified by your DAO.`,
+      variant: threshold === 1 ? 'secondary' : 'primaryPurple',
```
alexkeating, line 26: For some of these it may be good to abstract these different scenarios into configuration objects and have a top level conditional rather than having conditional for each param

libs/protocol-client/src/lib/graphql/queries.graphql
```diff
@@ -595,9 +595,17 @@ fragment ContributionFragment on Contribution {
   guilds {
     id
     guild_id
+    attestation_threshold
+    verification_status_id
+    verified
     guild {
       id
       name
+      verification_setting_id
```
alexkeating, line 604: Isn't verification_setting_id a duplicate of verification_setting.id

libs/protocol-client/src/lib/graphql/queries.graphql
```diff
@@ -595,9 +595,17 @@ fragment ContributionFragment on Contribution {
   guilds {
     id
     guild_id
+    attestation_threshold
+    verification_status_id
+    verified
```
alexkeating, line 600: This will be removed

apps/protocol-frontend/src/components/VerificationHover.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const verificationPropsMap = {
+    Verified: {
+      label:
+        'Feel free to attest to this contribution! Even though it has been verified for this DAO, additional attestations is good. ',
+      variant: 'tertiary',
+    },
+    Unverified: {
+      label:
+        threshold === 1
+          ? `This contribution only needs ${threshold} more attestation to be verified by your DAO! Be the one to push it through!`
```
alexkeating, line 24: I have this saying `null more attestation`

alexkeating: Also attestation may need to be plural in some situations

alexkeating: Also, all contributions have pill even if the dao does not have a verification framework

jonathanprozzi: I think that was part of the design -- at least as I understood it

libs/protocol-client/src/lib/graphql/queries.graphql
jonathanprozzi: What do you mean?

libs/protocol-client/src/lib/graphql/queries.graphql
jonathanprozzi: Ah, yes. I added this in before we switched that over

apps/protocol-frontend/src/components/VerificationHover.tsx
jonathanprozzi: Good catch on the `null` and the plurals -- not sure why I didn't see that locally since I tested with null and multiple attestations. Will need to look at that

jonathanprozzi: image

I think all the DAOs are pills. @twblack88 should this be?

alexkeating: If all the statuses are pills that is probably fine, but they have the message saying you are `x attestations away` so maybe we have a different message?

libs/protocol-client/src/lib/graphql/queries.graphql
jonathanprozzi: If this change is in #465 I'll fix the conflicts after that's in

libs/protocol-client/src/lib/graphql/queries.graphql
jonathanprozzi: Same as above -- if this change is in #465 I'll resolve this after it's merged

apps/protocol-frontend/src/components/VerificationHover.tsx
jonathanprozzi: I think I'm still unclear. My understanding is the status pill shows that message, and then the DAO column shows the checkmark on the DAO name only when it's verified.

The status pill should be showing either way -- is that not what's happening?

jonathanprozzi: attestation-table-vf-test

apps/protocol-frontend/src/components/AttestationsTable.tsx
jonathanprozzi: Good call, will work on that

libs/protocol-client/src/lib/graphql/queries.graphql
jonathanprozzi: I'll need to change this in a few other places as well outside of this PR

apps/protocol-frontend/src/components/VerificationHover.tsx
jonathanprozzi: Ok thanks, that helps narrow that down so I can replicate and handle the case better. I thought I tested  for both of those cases

apps/protocol-frontend/src/components/VerificationHover.tsx
jonathanprozzi: I made some changes to make the logic more readable but haven't moved it up into a configuration object yet -- wanted to make sure that the cases are all being covered.

I added the state we talked through earlier where a VF is set but attestation_threshold is null

jonathanprozzi:
@alexkeating I made the logic more readable and handled the case we talked through earlier. I haven't moved the logic into a configuration yet since I wanted to make sure everything is working properly first

alexkeating (approved):

PR #489, opened 2023-04-11
Chore/add back missing reverts

alexkeating:
## Description

- Add reverted changes back


jonathanprozzi:
Works well on create, but running into an error on any update (not only the activity type). I think maybe this happened before and was a permission issue?

image

Screencap:

update-activity-type


To replicate:

- Update a contribution

jonathanprozzi:
Looks good!

jonathanprozzi (approved):
All good now with the change, update and create both work!

PR #495, opened 2023-04-14
fix/create-more-dao-id-dropping (ENG-1047)

jonathanprozzi:
## Linear Ticket

- [ENG-1047](https://linear.app/govrn/issue/ENG-1047/bug-create-more-seems-to-be-dropping-the-dao-id)

## Description

- Noticed that when Create More is toggled on we were dropping the `daoId` value on the subsequent contribution

To test:

- Toggle on create more with a DAO added, and then add at least 1 more contribution and check that the DAO is applied to each contribution after the first


amrro (approved):
Good catch, and good fix. 

I also agree with deploying this directly to production.

jonathanprozzi:
I think at this point I'll merge into staging since we'll be merging into prod soon enough. Thanks!

PR #497, opened 2023-04-19
Feature/attestation threshold middleware

alexkeating:
## Description

- Add create middleware for threshold


apps/protocol-api/src/main.ts
```diff
@@ -18,6 +18,87 @@ import cors = require('cors');
...
+      },
+      where: {
+        id: guildContribution.id,
+      },
+    });
+  }
+};
+
+prisma.$use(async (params, next) => {
+  if ((params.model == 'Attestation', params.action === 'create')) {
```
jonathanprozzi, line 88: @alexkeating Is there where you mentioned that this middleware reads the params and then in this case if it's the Attestation model with the create action, then the middleware will run?

jonathanprozzi:
Also getting this error for creating contributions when I select a DAO

jonathanprozzi:
I'm getting the errors for most actions. I added a log to the middleware in the `prisma.$use` and a lot of the params are undefined, so wondering if this is causing some issues for our other actions?

apps/protocol-api/src/main.ts
jonathanprozzi, line 88: Changing this to `if ((params.model === 'Attestation' && params.action === 'create')) {`

fixes the issues I found. It looks like it was running this code path on anything with an action of `create` instead of only on Attestation && create

This is why I was seeing those undefined errors in the testing.

jonathanprozzi (changes requested):
Change to line 88 (see comment) -- I need to get in the habit of using the 'Request Changes' instead of only 'Comment'

PR #498, opened 2023-04-19
add canonical type table + migrated from field

Flip-Liquid:
## Linear Ticket

[1043](https://linear.app/govrn/issue/ENG-1043/implement-table-schema)

## Description

Add canonical type table, and migrated from field in existing guild activity type table.

## Screencaptures


apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -285,16 +301,18 @@ model ContributionStatus {
...
+  updatedAt         DateTime               @default(now()) @updatedAt
+  name              String                 @unique
+  active            Boolean                @default(true)
+  default           Boolean                @default(false)
+  users             UserActivity[]
+  contributions     Contribution[]
+  categoryActivity  CategoryActivityType[]
+  guilds            GuildActivityType[]
+  GuildActivityType GuildActivityType[]
+  GuildActivityType GuildActivityType[]
```
jonathanprozzi, line 315: This duplication is causing an error in the build -- Should one of these be the `CanonicalGuildActivityType`?

Flip-Liquid:
Yeah ir was auto-included by a litter I have installed, need to address.
Was more just submitting this to review the approach to the schema change;
definitely still a WIP

jonathanprozzi:
oh got it! sounds good thanks!

jonathanprozzi:
So my understanding of this is that the canonical are whichever activity types are currently decided as being active (canonical) by the guild, and that the migrated from field is how we'll be handling any edits/updates?

Flip-Liquid:
> So my understanding of this is that the canonical are whichever activity types are currently decided as being active (canonical) by the guild, and that the migrated from field is how we'll be handling any edits/updates?
> 
> Does that mean that we'll want to switch anything that relates to activity types as they are currently to use the canonical table instead?

Anywhere we want to check to see or update what the canonical types for a guild are, we'll be hitting that table. We'll still need to join against the Activity table, since that's where the actual data for the type are.

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -142,17 +143,33 @@ model GuildContributionVerificationStatus {
...
-  guild            Guild        @relation(fields: [guild_id], references: [id])
-  activity_type_id Int
-  activity_type    ActivityType @relation(fields: [activity_type_id], references: [id])
+  id                         Int                          @id @default(autoincrement())
+  createdAt                  DateTime                     @default(now())
+  updatedAt                  DateTime                     @default(now()) @updatedAt
+  guild_id                   Int
+  guild                      Guild                        @relation(fields: [guild_id], references: [id])
+  activity_type_id           Int
+  activity_type              ActivityType                 @relation(name: "guilds", fields: [activity_type_id], references: [id])
```
amrro, line 152: the custom name guilds doesn't seem to accurately describe the relation between the columns, as it is a foreign key to activity types. Please check my other comments.

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -75,22 +75,23 @@ model DiscordUser {
...
+  status                         GuildStatus                  @default(VALIDATED)
   contributions                  GuildContribution[]
   users                          GuildUser[]
   twitter_accounts               TwitterAccount[]
   activity_type                  GuildActivityType[]
   contribution_reporting_channel String?
   guild_imports                  GuildImport[]
   verification_setting_id        Int?
-  verification_setting           VerificationSetting? @relation(fields: [verification_setting_id], references: [id])
+  verification_setting           VerificationSetting?         @relation(fields: [verification_setting_id], references: [id])
```
amrro, line 93: This approach to declaring foreign key relations is consistent with the conventions used throughout the rest of the project.

amrro (changes requested):
It seems PR is incomplete. You are going to need to include the missing files: 
- ts generated prisma client files. 
- migrations sql files.

To generate these files, run: 
```console
yarn prisma migrate dev --schema=<path-to-schema>
``` 

Also, as a sanity check, it's a good idea to run the generating and linting jobs on `protocol-client` to ensure everything is in order..

I've provided some feedback for the included file in the PR.

Flip-Liquid:
> It seems PR is incomplete. You are going to need to include the missing files:
> 
> * ts generated prisma client files.
> * migrations sql files.
> 
> Including only the bare schema changes in the staging environment is not recommended IMHO, as it could block other developments and mess with git operations.
> 
> Also, as a sanity check, it's a good idea to run the generating and linting jobs on `protocol-client` to ensure everything is in order..
> 
> I've provided some feedback for the included file in the PR.

Yep. this is intentional. I didn't generate the files since I want to close on the actual schema first, and to minimize delta until we've settled on the schema.

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -142,17 +143,33 @@ model GuildContributionVerificationStatus {
...
-  activity_type    ActivityType @relation(fields: [activity_type_id], references: [id])
+  id                         Int                          @id @default(autoincrement())
+  createdAt                  DateTime                     @default(now())
+  updatedAt                  DateTime                     @default(now()) @updatedAt
+  guild_id                   Int
+  guild                      Guild                        @relation(fields: [guild_id], references: [id])
+  activity_type_id           Int
+  activity_type              ActivityType                 @relation(name: "guilds", fields: [activity_type_id], references: [id])
+  migrated_from_id           Int?
+  migrated_from              ActivityType?                @relation(name: "migrated_from", fields: [migrated_from_id], references: [id])
```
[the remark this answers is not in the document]
Flip-Liquid: In this case, we need to define custom names, since we have two relations referencing the same table. both are from `GuildActivityType` -> `ActivityType`, and the prisma linter considers it an error to not name the relations manually. although we could probably just name the second/new relation as opposed to both.

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -294,7 +311,8 @@ model ActivityType {
   users            UserActivity[]
   contributions    Contribution[]
   categoryActivity CategoryActivityType[]
-  guilds           GuildActivityType[]
+  guilds           GuildActivityType[]    @relation(name: "guilds")
```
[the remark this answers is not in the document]
Flip-Liquid: explicit management here is required. It might have changed when we upgraded to prisma 3, but there were implicit relations allowed in prisma 2 that are now disallowed in 3.

specifically, prisma 3 is not allowing two different fk references from GuildActivityType to ActivityType. See the below error:

image

Flip-Liquid:
@amrro replied to your comments. could you please re-review with those in mind? Like I mentioned, the autogenerated code is intentionally omitted until we close on the above schema. I'll send an iteration on this PR with those code changes once we agree on the schema. specifically, could you please review the semantic changes to the schema:

1. do you agree with the proposed fields in CanonicalGuildActivityType?
2. Do you agree with the proposed structure of the migrated_from field in GuildActivityType?

apps/protocol-api/src/prisma/schema.prisma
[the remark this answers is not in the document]
amrro: Ahaa. Thank you for bringing this up.. I missed this point.

apps/protocol-api/src/prisma/schema.prisma
amrro: In this case, using descriptive name that reflects the relation between two models ensures readability for constraints names. For examples, using a more specific name than "guilds" would definitely help to avoid confusion.

amrro:
Thanks for providing more context. Your comments have been very helpful in resetting the discussion.

> 1. do you agree with the proposed fields in CanonicalGuildActivityType?
> 2. Do you agree with the proposed structure of the migrated_from field in GuildActivityType?


Here is my understating to the schema update is that a new table is going to be added called, `CanonicalGuildActivityTypes`. This table has the following relations: 
1. one-to-many relation with `GuildActivityType`. 
2. one-to-many relation with `Guild`<sup>!?</sup>.

However, I have some concerns about the relation between `Guild` and `CanonicalGuildActivityTypes`. In my opinion, adding this relation could potentially add more complexity and may affect the existing codebase or API usage. 

Instead, I suggest a simpler approach of replacing the `Guild` relation with a one-to-many relation between `ActivityType` (single source of truth) and `CanonicalGuildActivityTypes`. This would remove the metadata that exists on `CanonicalGuildActivityTypes` like "name" and simplify the schema. And this can lead to a simpler represenation for mapping b/w activity types. In addition, `Guild`'s API would be as same as previous.

Here is my visual summary: 
IMG_86096FC43C42-1



Please let me know your thoughts on this approach.


> the autogenerated code is intentionally omitted until we close on the above schema. I'll send an iteration on this PR with those code changes once we agree on the schema

Alternatively, You could have multiple PRs per one migration file. My concern is that merging this could block any further updates to the schema.

apps/protocol-api/src/prisma/schema.prisma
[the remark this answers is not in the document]
amrro: I guess this crucial to let prisma know what's type of this relation and for generated ts files. 

I would like to ask. Do we need to limit the functionality of the opposite direction queries?

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -75,22 +75,23 @@ model DiscordUser {
...
   contributions                  GuildContribution[]
   users                          GuildUser[]
   twitter_accounts               TwitterAccount[]
   activity_type                  GuildActivityType[]
   contribution_reporting_channel String?
   guild_imports                  GuildImport[]
   verification_setting_id        Int?
-  verification_setting           VerificationSetting? @relation(fields: [verification_setting_id], references: [id])
+  verification_setting           VerificationSetting?         @relation(fields: [verification_setting_id], references: [id])
+  CanonicalGuildActivityType     CanonicalGuildActivityType[]
```
amrro, line 94: Prisma format uses CamelCase format, But we are using snake. 

I remember an igniting discussion about if we need to change this.

Flip-Liquid:
Thanks for the feedback. Replying to your comment,

> > the autogenerated code is intentionally omitted until we close on the above schema. I'll send an iteration on this PR with those code changes once we agree on the schema
> 
> Alternatively, You could have multiple PRs per one migration file before going to main. My concern is that merging this could block any further updates to the schema.

Again, I'm not planning on merging this as is. once we have consensus on the schema changes, I'll add the autogenerated code to this PR and then we can merge that in. I'm not sure how this would block updating the schema.

> Here is my understating to the schema update is that a new table is going to be added called, `CanonicalGuildActivityTypes`. This table has the following relations:
> 
> 1. one-to-many relation with `GuildActivityType`.
> 2. one-to-many relation with `Guild`!?.

on the relations, the intention is the following:
One canonical guild activity type can refer to one guild and one guild activity type. (satisfied by the `@@unique` clause)
One guild can be referred to by many guild activity types, and many canonical guild activity types. (explicit in the schema)
One guild activity type can be referred to by one canonical guild activity type. (just updated this to ensure uniqueness of the guild_activity_id in the canonical table)

> However, I have some concerns about the relation between `Guild` and `CanonicalGuildActivityTypes`. In my opinion, adding this relation could potentially add more complexity and may affect the existing codebase or API usage.
> 
> Instead, I suggest a simpler approach of replacing the `Guild` relation with a one-to-many relation between `ActivityType` (single source of truth) and `CanonicalGuildActivityTypes`. This would remove the metadata that exists on `CanonicalGuildActivityTypes` like "name" and simplify the schema. And this can lead to a simpler represenation for mapping b/w activity types. In addition, `Guild`'s API would be as same as previous.
> 
> Here is a visual summary: IMG_86096FC43C42-1
> 

I'm not quite seeing how this is simpler than the proposed approach. I think the 'name' field on CanonicalGuildActivity type could be eliminated either way, since we would ultimately take the same name as the underlying activity type (@jonathanprozzi 
 about removing 'name' also). I also think having a list of the references on the guild of the canonical types is advantageous e.g. for displaying the data on the DAO dashboard. Likewise, Having the `guild_id` field in `CanonicalGuildActivityType` field lets us filter that table on a particular Guild's id to easily find the list of canonical types for a given guild. 

For those operations, at least, I think removing the reference to `Guild` from the canonical types would make things more complex. Let me know if that addresses what you're thinking @amrro .

amrro:
> > > the autogenerated code is intentionally omitted until we close on the above schema. I'll send an iteration on this PR with those code changes once we agree on the schema
> > 
> > 
> > Alternatively, You could have multiple PRs per one migration file before going to main. My concern is that merging this could block any further updates to the schema.
> 
> Again, I'm not planning on merging this as is. once we have consensus on the schema changes, I'll add the autogenerated code to this PR and then we can merge that in. I'm not sure how this would block updating the schema.

Apologies for the confusion. 



> One canonical guild activity type can refer to one guild and one guild activity type. (satisfied by the `@@unique` clause)
> One guild can be referred to by many guild activity types, and many canonical guild activity types. (explicit in the schema)
> One guild activity type can be referred to by one canonical guild activity type. (just updated this to ensure uniqueness of the guild_activity_id in the canonical table)

Okay, that's make sense.

I think all is left is some minor changes.

apps/protocol-api/src/prisma/schema.prisma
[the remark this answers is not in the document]
Flip-Liquid: I don't think so. i think this is a convenience field. but I don't fully understand how this works with prisma, i admit. I propose we take any functionality limiting we need to do as a separate PR.

apps/protocol-api/src/prisma/schema.prisma
Flip-Liquid: updated to use snake_case

amrro (approved):
Thank you for your patience throughout the review process.
Great work on your end as well!

PR #503, opened 2023-04-26
qa/verification-framework-ui-change (ENG-1064) [WIP]

jonathanprozzi:
## Linear Ticket

- [ENG-1064](https://linear.app/govrn/issue/ENG-1064/vf-changes-prior-to-release)

## Description

## Screencaptures

### DIFF
--- a/apps/protocol-frontend/src/components/MyAttestationsTable.tsx
+++ b/apps/protocol-frontend/src/components/MyAttestationsTable.tsx
@@ -72,7 +72,7 @@ const MyAttestationsTable = ({
         },
       },
       {
-        header: 'Status',
+        header: 'Verification',
         accessorFn: contribution =>
           contribution.guilds[0]?.verificationStatus?.name === 'Verified'
             ? 'Verified'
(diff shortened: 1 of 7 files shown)


apps/protocol-frontend/src/components/MyAttestationsTable.tsx
```diff
@@ -86,21 +86,30 @@ const MyAttestationsTable = ({
...
 
+          const guildHasVerificationFramework =
+            row.original.guilds[0].guild?.verification_setting_id !== null;
+          const attestationThreshold =
+            row.original.guilds[0].attestation_threshold;
+
+          const frameworkSettingThreshold =
+            row.original.guilds[0].guild.verification_setting
+              ?.num_of_attestations;
+          const pillUnverifiedLabel = `${attestationThreshold}/${frameworkSettingThreshold}`;
```
Flip-Liquid, line 97: these are usually gonna be the same (row.original.guilds[0].attestation_threshold; & row.original.guilds[0].guild.verification_setting?.num_of_attestations;) since the latter is copied onto the former each time the job runs. if you want # attestations / # attestations need for verified, you should get the number of current attestations on the contribution.

Flip-Liquid (changes requested):
small comment on the code, otherwise looks good

apps/protocol-frontend/src/components/MyAttestationsTable.tsx
jonathanprozzi: Thanks! I think this may fix the issue I'm facing

Flip-Liquid (approved):
thanks jp

PR #507, opened 2023-04-27
Refactor IPFS URL to HTTP URL

amrro:
## Linear Ticket

[ENG-1020](https://linear.app/govrn/issue/ENG-1020/bug-view-proof-link-broken)

## Description

Introduce a utility function `ipfsToHttp` to validate and convert IPFS URLs to their corresponding HTTP URLs by extracting the CID and generating the appropriate HTTP URL.


jonathanprozzi (approved):
Awesome! 
One thing I'm wondering about and we may need to follow up with @alexkeating or @amsos is if we have a gateway on our account.

jonathanprozzi (commented):
Nonblocking question -- do we want to move the `GATEWAY_URL` to our env instead?

PR #514, opened 2023-05-01
Reposition Personal Types in dropdown Menus for Report & Edit Contribution Forms

amrro:
## Linear Ticket

[ENG-1058 - Move personal types to the top of the dropdown](https://linear.app/govrn/issue/ENG-1058/move-personal-types-to-the-top-of-the-dropdown)

## Description
This PR includes the following updates:
- In both Edit and Report Contribution Forms, personal types have been moved to the top of the dropdown menus for improved accessibility.
- A "Default" label has been added to the default options in the dropdowns to clearly distinguish them from other options.

### DIFF
--- a/apps/protocol-frontend/src/components/EditContributionForm.tsx
+++ b/apps/protocol-frontend/src/components/EditContributionForm.tsx
@@ -153,7 +153,7 @@ const EditContributionForm = ({ contribution }: EditContributionFormProps) => {
       })),
     ];
     if (personalTypes) {
-      results.push(personalTypes);
+      return [personalTypes, ...results];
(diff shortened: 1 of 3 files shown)


Flip-Liquid:
@amrro looks good to me if we've had the discussion about the ordering (i.e. what order do we want DAO/default/personal activity types in?)

jonathanprozzi (approved):
Looks great. I agree with your question that we may want to think about how we want to label the Default types in the future, but nice job! Definitely a UX improvement.

🎉

Let's hold off on merging this in until we get all the VF feature out the door

PR #515, opened 2023-05-01
Handle empty attestation state for users with no contributions to attest

amrro:
## Linear Ticket

[ENG-1009 - Attestations Empty state is triggering](https://linear.app/govrn/issue/ENG-1009/attestations-empty-state-is-triggering)

## Description

Previously, an incorrect message was displayed when a user had joined one or more DAOs but had no contributions to attest. This PR fixes it by providing a clearer message and CTA for this case.

## Screencaptures
Screenshot 2023-05-01 at 5 58 22 PM


jonathanprozzi:
Nice job! This looks great to me, but going to also tag @christinevandev for a quick copy review.

jonathanprozzi (approved):
Approving this pending any minor copy feedback from Christine, but looks great! 🎉

(Let's hold off on merging until after the VF release to minimize moving pieces in QA)

PR #521, opened 2023-05-04
fix/middleware-logic

jonathanprozzi:
## Linear Ticket

## Description

- Relates to error in the Linear job that uses a `.createMany` call and could be causing unexpected behavior

## Screencaptures

### DIFF
--- a/apps/protocol-api/src/main.ts
+++ b/apps/protocol-api/src/main.ts
@@ -93,7 +93,7 @@ prisma.$use(async (params, next) => {
       chain_id: contributionParams.chain_id,
     });
   }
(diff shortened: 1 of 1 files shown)


apps/protocol-api/src/main.ts
```diff
@@ -93,7 +93,7 @@ prisma.$use(async (params, next) => {
       chain_id: contributionParams.chain_id,
     });
   }
-  if ((params.model == 'Attestation', params.action === 'createMany')) {
```
Flip-Liquid, line 96: great catch. what was this evaluating to beforehand? what is the truth value of (bool, bool)?

Flip-Liquid (approved):

PR #523, opened 2023-05-09
logging utils

Flip-Liquid:
## Linear Ticket

[eng-1078](https://linear.app/govrn/issue/ENG-1078/logs-metrics-instrumentation-for-contract-sync-job)

## Description

move to json logging, augment logging for contract sync job

## Screencaptures

### DIFF
--- a/apps/contract-sync-job/src/main.ts
+++ b/apps/contract-sync-job/src/main.ts
@@ -12,6 +12,7 @@ import {
   upsertAttestation,
 } from './db';
 import { batch, MintedContributionSchemaV1 } from '@govrn/protocol-client';
+import { getLogger } from '@govrn/utils';
 
 const CHAIN_NAME = process.env.CHAIN_NAME;
 const JOB_NAME = `contract-sync-job-${CHAIN_NAME}`;
@@ -30,9 +31,10 @@ const networkConfig: NetworkConfig = {
 
 const provider = new ethers.providers.JsonRpcProvider(CHAIN_URL);
 const govrnContract = new GovrnContract(networkConfig, provider);
+export const logger = getLogger('contract-sync-job', CHAIN_NAME);
 
 const main = async () => {
-  console.log(':: Starting to Process Contribution(s)');
+  logger.info(':: Starting to Process Contribution(s)');
 
   const client = new GovrnGraphClient(CHAIN_ID);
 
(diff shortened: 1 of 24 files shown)


libs/utils/src/lib/logging.ts
```diff
@@ -0,0 +1,14 @@
+import { createLogger, format, transports } from 'winston';
+
+const { timestamp, json, colorize } = format;
+
+// TODO: cust class appending appending more info per log
+export const getLogger = (source: string, tag: string) => {
+  const logger = createLogger({
+    level: 'info',
+    format: format.combine(timestamp(), colorize(), json()),
```
amrro, line 9: It appears that the ANSI color codes are not rendering correctly, which hinder readability. The raw ANSI codes are shwon: 

Screenshot 2023-05-09 at 12 38 24 PM

Flip-Liquid: that's because it's being output as JSON for the benfit of datadog. for now, I'm just going to drop colorization.

apps/contract-sync-job/src/main.ts
```diff
@@ -115,34 +126,34 @@ const main = async () => {
       async contribution => await upsertContribution(contribution),
       BATCH_SIZE,
     );
-    console.error('batch Contribution errors', errors);
+    logger.error(':: BATCH CONTRIBUTION ERRORS', errors);
 
     const upsertedCount = inserted.length;
     if (upsertedCount > 0) {
-      console.log(`:: Inserted ${upsertedCount} Contribution(s)`);
+      logger.info(`:: Inserted ${upsertedCount} Contribution(s)`);
```
amrro, line 133: Do you think that `::` suffix is still necessary?

libs/utils/src/lib/logging.ts
```diff
@@ -0,0 +1,14 @@
+import { createLogger, format, transports } from 'winston';
+
+const { timestamp, json } = format;
+
+// TODO: cust class appending appending more info per log
```
amrro, line 5: This todo could be replaced with a linear issue instead.

amrro (approved):
GREAT! LGTM.

There is only couple of comments.

apps/contract-sync-job/src/main.ts
Flip-Liquid: no, probably not, i can remove it in an additional PR if you like

PR #531, opened 2023-05-15
feature/dashboard-hover (ENG-1031)

jonathanprozzi:
## Linear Ticket

- [ENG-1031](https://linear.app/govrn/issue/ENG-1031/user-dashboard-show-contributions-on-hover)

## Description

- Adds a hover to the dashboard that is similar to the one on the DAO dashboard
- Shows the `name - DAO` with `Unassigned` showing for unattributed contributions
- Shows the 5 most recent, and then if there are more than 5 it shows `... and X more!`

## Screencaptures

dashboard-hover-new


apps/protocol-frontend/src/components/ContributionsHeatMap.tsx
```diff
@@ -40,7 +121,7 @@ const ContributionsHeatMap = ({
             data={contributionsCountMap.filter(
               contribution => contribution.value !== 0,
             )}
-            from={subWeeks(new Date(), 52)}
+            from={subWeeks(new Date(), YEAR)}
```
Flip-Liquid, line 124: nice. nit: maybe YEAR_WEEKS?

Flip-Liquid (approved):

apps/protocol-frontend/src/components/ContributionsHeatMap.tsx
jonathanprozzi: great call. updated that + the impacted files

PR #537, opened 2023-05-17
add twitter integration field to guild model

Flip-Liquid:
## Linear Ticket

[1107](https://linear.app/govrn/issue/ENG-1107/backend-schema-updates-for-t2c)

## Description

Prisma schema changes to support Guild Twitter integrations

## Screencaptures


apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -157,6 +159,15 @@ model GuildActivityType {
   @@unique([guild_id, activity_type_id])
 }
 
+model GuildTwitterIntegration {
+  id        Int      @id @default(autoincrement())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+  guild_id  Int      @unique
+  Guild     Guild    @relation(fields: [guild_id], references: [id])
+  hashtag   String   @unique
```
jonathanprozzi, line 168: definitely great call to start with this unique, and we can always relax it a bit if we decide to

jonathanprozzi (approved):

PR #540, opened 2023-05-22
bug/improve-activity-types-invalidation-ux (ENG-1018)

jonathanprozzi:
## Linear Ticket

- [ENG-1018](https://linear.app/govrn/issue/ENG-1018/bug-improve-activity-types-invalidation-ux)

## Description

- Changes the invalidation for the contribution mutations (create and update) to invalidate the user's personal activity types query regardless of the args supplied
  - This enables the user to utilize the newly created activity type in the 'Create more' flow if they clear it and want to reselect it. It now behaves similarly to other data fields
- Makes the Activity Types selects now clearable via the X and the backspace (same as the DAO select) on both the create and edit contribution forms 

## Screencaptures

https://github.com/Govrn-HQ/govrn-monorepo/assets/9438776/d0f0c528-de05-4b8d-817a-85ca584f5e1a


amrro (approved):
LGTM. This adds up to UX. 

It's working just fine. But I wonder how changing query's key fix it for Create more option. I might missing something.

jonathanprozzi:
> LGTM. This adds up to UX.
> 
> It's working just fine. But I wonder how changing query's key fix it for Create more option. I might missing something.

Oh, it's not directly related to the Create More functionality. That was an example of where having the background invalidation would allow for someone to switch since they won't be closing the report form / reloading the page.

PR #542, opened 2023-05-25
Add  `<RequireAuth />` component for Authentication

amrro:
## Linear Ticket

[ENG-1030 - Improve Disconnected User UX](https://linear.app/govrn/issue/ENG-1030/improve-disconnected-user-ux)

## Description

`<RequireAuth />`` component manages authentication for wrapped pages in it. This component is now responsible for determining user authentication status and providing option to connect wallet without re-routing.

## Screencaptures

Screenshot 2023-06-01 at 7 45 38 PM


apps/protocol-frontend/src/components/ConnectWallet.tsx
```diff
@@ -12,16 +12,20 @@ import {
...
   Box,
   Flex,
+  ButtonProps,
 } from '@chakra-ui/react';
 import { FiKey, FiChevronDown, FiXCircle } from 'react-icons/fi';
 import { useEnsNameEthers } from '../hooks/useEnsNameEthers';
 import { displayAddress } from '../utils/web3';
 
-interface ConnectWalletProps {
+interface ConnectWalletProps extends ButtonProps {
```
jonathanprozzi, line 21: good idea!

apps/protocol-frontend/src/utils/requireAuth.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+type RequireAuthProps = {
+  children: React.ReactNode;
+};
+
+const NonAuthenticatedView = () => {
+  return (
+    <GovrnCta
+      heading={'Welcome to Govrn'}
+      emoji={'🫡'}
```
jonathanprozzi, line 16: I'm thinking that the particular emoji we used in the Figma may not be fully supported.

We can replace it with 👋 - Should be able to copy this directly from the one in the Dashboard.tsx

jonathanprozzi (commented):
This looks great and streamlines quite a lot. Only question is around the rendering of the emoji. I think the one we selected may not be fully supported so I suggest we replace with 👋 for now.

jonathanprozzi:
Looks great. Emoji working better now!

One question for everyone (and not part of the scope here but the PR changes made this more apparent)

What do we think about using a primary button instead of tertiary in the Sidebar similar to what @amrro has in place for this component? I really love how the primary looks on the component but wondering if it's too much to be in the Sidebar also?

image

This isn't something we need to think about for this PR, but curious since it would be a quick change to make the ConnectWallet be primary in the Sidebar and MobileNav

I like the Connect Wallet being primary in the Sidebar but wondering if it's too much? That was the original reason we didn't use it. It looks great in the new component so figured I'd surface again!

@amsos @christinevandev

jonathanprozzi (approved):
Great job! My comment about the Sidebar doesn't block merge at all, mainly wanted to surface that since it looks really good in the new component.

amrro:
> What do we think about using a primary button instead of tertiary in the Sidebar similar to what @amrro has in place for this component? I really love how the primary looks on the component but wondering if it's too much to be in the Sidebar also?

I agree with you. Having both of the primary buttons showing might be visually overwhelming or distracting. a tertiary style for the sidebar might be more balanced without overloading the interface, I think so, @jonathanprozzi.

PR #547, opened 2023-05-30
fix/update-shield-permissions

jonathanprozzi:
## Linear Ticket

n/a

## Description

- Removing `twitter_integration_id` from the shield perms since we removed it in #543 

## Screencaptures

n/a


Flip-Liquid (approved):
thanks jp. I'll take a look as to why this was not caught in CI

PR #557, opened 2023-06-08
added user check by address back in to flow

jonathanprozzi:
## Linear Ticket

- loosely relates to ENG-939

## Description

- Quick fix to add a check back in -- more fully resolved in upcoming rework


Flip-Liquid (approved):
thanks for catching this @jonathanprozzi

PR #558, opened 2023-06-08
twitter tweet id type

Flip-Liquid:
## Linear Ticket

na
## Description

schema and app fixes for changing the unique identifier of a twitter tweet

## Screencaptures


apps/protocol-api/src/main.ts
```diff
@@ -338,4 +338,5 @@ app.get('/discord/oauth', async function (req, res) {
 });
 
 console.log('listening');
+console.log(process.env.NODE_ENV);
```
jonathanprozzi, line 341: did you add this for debugging purposes? just curious

Flip-Liquid: yep. was running aground on shield issues and wanted to make sure the proper env was being set

jonathanprozzi (approved):

PR #561, opened 2023-06-13
eng-1133/qa-join-dao-view

jonathanprozzi:
## Linear Ticket

- [ENG-1133](https://linear.app/govrn/issue/ENG-1133/qa-join-dao-view)

Note: Putting this up for review but we may want to make minor copy changes prior to merge.

## Description

- Checks if DAO exists and displays alternate CTA if not
- Resolves issue we saw with the Join DAO from this particular DAO route

## Screencaptures

https://github.com/Govrn-HQ/govrn-monorepo/assets/9438776/1616eaef-987a-432b-bbee-b3fabd2ca696


Flip-Liquid (approved):
nice fixup

PR #580, opened 2023-06-20
tip schema

Flip-Liquid:
## Linear Ticket

[ENG-1139](https://linear.app/govrn/issue/ENG-1139/tipping-backend-schema)

## Description

schema changes for tippin'

## Screencaptures


apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -19,16 +19,16 @@ enum GuildStatus {
 }
 
 model User {
```
jonathanprozzi, line 21: Nice -- did we also decide we wanted to do a `tipping_enabled` boolean so we have the option to use that to check enabled/disabled instead of the contract address only? Thinking in a situation where we've created the split, a user disables and then reenables. We could set the true/false of `tipping_enabled` so we can use the original contract address

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -38,6 +38,21 @@ model User {
   guild_users          GuildUser[]
   twitter_user         TwitterUser?
   discord_users        DiscordUser[]
+  tips                 Tip[]
+  split_contracts      UserSplitContract[]
```
jonathanprozzi, line 42: Good idea! Gives us leeway to support multiple per user!

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -38,6 +38,21 @@ model User {
...
+
+model UserSplitContract {
+  id        Int      @id @default(autoincrement())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+  user_id   Int
+  user      User     @relation(fields: [user_id], references: [id])
+  chain_id  Int?
+  chain     Chain?   @relation(fields: [chain_id], references: [id])
+  address   String
```
jonathanprozzi, line 53: This would be the address of the contract? Just double checking

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -221,12 +236,13 @@ model ChainType {
...
-  name          String         @unique
-  chain_id      String         @unique
-  contributions Contribution[]
+  id                   Int                 @id @default(autoincrement())
+  createdAt            DateTime            @default(now())
+  updatedAt            DateTime            @default(now()) @updatedAt
+  name                 String              @unique
+  chain_id             String              @unique
+  contributions        Contribution[]
+  user_split_contracts UserSplitContract[]
```
jonathanprozzi, line 245: Great idea!

apps/protocol-api/src/prisma/schema.prisma
```diff
@@ -458,6 +474,22 @@ model JobRun {
   name          String
 }
 
+////
+// Tips model
+////
+
+model Tip {
```
jonathanprozzi, line 481: Looks great! Would this include both the user (tippee) and the tipper?
I know we talked about that at one point but wasn't exactly sure where we landed

jonathanprozzi (approved):
A few clarifying questions on my part, but looks great

apps/protocol-api/src/prisma/schema.prisma
Flip-Liquid: yes! great catch. sorry i missed that and thank you for doing a thorough review.

apps/protocol-api/src/prisma/schema.prisma
Flip-Liquid: correct!

apps/protocol-api/src/prisma/schema.prisma
Flip-Liquid: yep. idea is to support multiple chains.

jonathanprozzi:
Looks great!

apps/protocol-api/src/prisma/schema.prisma
jonathanprozzi: Absolutely! Great job!

